PR #20, opened 2017-06-30
EligibilityCheckController now only takes first two characters of cou…

AndyHWChung:
…ntry code from user info API


AndyHWChung:
c.f https://jira.tools.tax.service.gov.uk/browse/PE-3061

PR #28, opened 2017-08-11
ITMP interface updated

AndyHWChung:


AndyHWChung:
see https://drive.google.com/file/d/0B6ofRdAnJAwgeG5GRlpXcTljc3c/view

AndyHWChung:
I've also separated out the eligibilty checks and the retrieval of user information into two separate endpoints - getting user information won't be done in the backend soon so the separation makes it easier to get rid of it from our code

PR #42, opened 2017-09-26
HTS-479 : Library upgrade and added auth-client to secure endpoints

sureshhmrc:


app/uk/gov/hmrc/helptosave/controllers/HelpToSaveAuth.scala
```diff
@@ -0,0 +1,77 @@
...
+        .retrieve(Retrievals.authorisedEnrolments) { authorisedEnrols ⇒
+
+          val mayBeNino = authorisedEnrols.enrolments
+            .find(_.key === "HMRC-NI")
+            .flatMap(_.getIdentifier("NINO"))
+            .map(_.value)
+
+          mayBeNino.fold(
+            toFuture(InternalServerError("could not find NINO for logged in user"))
+          )(_ ⇒ action(request))
```
AndyHWChung: We probably don't need this right now, but might it be useful to have `action` depend on the NINO as well so things calling `authorised` can have access to the NINO if needed? We can then eliminate the need for the NINO being passed in through the backend endpoints.

app/uk/gov/hmrc/helptosave/controllers/HelpToSaveAuth.scala
```diff
@@ -0,0 +1,77 @@
...
+        }.recover {
+          handleFailure()
+        }
+    }
+
+  def handleFailure(): PartialFunction[Throwable, Result] = {
+    case _: NoActiveSession ⇒
+      logger.warn("user is not logged in, probably a hack? ")
+      InternalServerError("")
```
AndyHWChung: would passing in a descriptive string to `InternalServerError` be helpful for debugging?

test/uk/gov/hmrc/helptosave/controllers/AuthSupport.scala
```diff
@@ -0,0 +1,57 @@
...
+
+import scala.concurrent.{ExecutionContext, Future}
+
+trait AuthSupport extends TestSupport {
+
+  val NinoWithCL200: Enrolment = Enrolment("HMRC-NI").withConfidenceLevel(L200)
+
+  val AuthProvider: AuthProviders = AuthProviders(GovernmentGateway)
+
+  val AuthWithCL200: Predicate = NinoWithCL200 and AuthProvider
```
AndyHWChung: this and the variables above are the same as the ones in `HelpToSaveAuth` - is it worth putting them in some object like the companion object of `HelpToSaveAuth` to avoid duplication?

test/uk/gov/hmrc/helptosave/controllers/HelpToSaveAuthSpec.scala
```diff
@@ -0,0 +1,73 @@
...
+    "handle NoActiveSession exception and throw an error" in {
+
+      val exceptions = List(
+        "BearerTokenExpired",
+        "MissingBearerToken",
+        "InvalidBearerToken",
+        "SessionRecordNotFound",
+        "InsufficientEnrolments",
+        "InsufficientConfidenceLevel",
+        "UnsupportedCredentialRole")
```
AndyHWChung: auth-client's 'fromString` also takes in the case where the string is something unexpected:

https://github.com/hmrc/auth-client/blob/master/src/main/scala/uk/gov/hmrc/auth/core/exceptions.scala#L56

To bee complete we should probably test this case as well

app/uk/gov/hmrc/helptosave/config/microserviceWiring.scala
```diff
@@ -55,7 +60,7 @@ trait WSHttp
 @Singleton
 class WSHttpExtension extends WSHttp with HttpAuditing with ServicesConfig {
 
-  override val hooks: Seq[HttpHook] = Seq(AuditingHook)
+  override val hooks: Seq[HttpHook] = NoneRequired
```
AndyHWChung: does this mean we're not doing auditing in the backend?

app/uk/gov/hmrc/helptosave/controllers/HelpToSaveAuth.scala
sureshhmrc: yeah, all endpoints require the nino, so we can derive it from here instead of passing it from frontend - but in a different PR  ?

app/uk/gov/hmrc/helptosave/config/microserviceWiring.scala
sureshhmrc: yeah, auditing between the service to service calls are not needed

test/uk/gov/hmrc/helptosave/controllers/HelpToSaveAuthSpec.scala
sureshhmrc: yeah, i was lazy -  i can add it now

app/uk/gov/hmrc/helptosave/controllers/HelpToSaveAuth.scala
AndyHWChung: yeah, that'd be better. Doing that would involve changing the frontend code as well

sureshhmrc: fixed

app/uk/gov/hmrc/helptosave/controllers/HelpToSaveAuth.scala
sureshhmrc: fixed

test/uk/gov/hmrc/helptosave/controllers/AuthSupport.scala
sureshhmrc: fixed

conf/app.routes
```diff
@@ -1,8 +1,8 @@
 # microservice specific routes
 
-GET /eligibility-check       @uk.gov.hmrc.helptosave.controllers.EligibilityCheckController.eligibilityCheck(nino: String)
-GET /enrolment-status        @uk.gov.hmrc.helptosave.controllers.EnrolmentStoreController.getEnrolmentStatus(nino: String)
-GET /enrol-user              @uk.gov.hmrc.helptosave.controllers.EnrolmentStoreController.enrol(nino: String)
-GET /set-itmp-flag           @uk.gov.hmrc.helptosave.controllers.EnrolmentStoreController.setITMPFlag(nino: String)
-GET /store-email             @uk.gov.hmrc.helptosave.controllers.EmailStoreController.store(email: String, nino: String)
-GET /get-email               @uk.gov.hmrc.helptosave.controllers.EmailStoreController.get(nino: String)
+GET /eligibility-check       @uk.gov.hmrc.helptosave.controllers.EligibilityCheckController.eligibilityCheck()
```
AndyHWChung: putting this change here now means that we need to update the frontend at the same time otherwise the master branches of FE and BE will not be able to communicate with each other

sureshhmrc: yes, getting the frontend ready

sureshhmrc: https://github.com/hmrc/help-to-save-frontend/pull/219

AndyHWChung:
LGTM

PR #45, opened 2017-10-03
Updated set ITMP flag interface

AndyHWChung:


AndyHWChung:
https://jira.tools.tax.service.gov.uk/browse/HTS-75

PR #56, opened 2017-11-02
HTS-563 : Increase Play request timeout from 20s to 43s

tim-holbrook:


conf/application.conf
```diff
@@ -126,6 +126,8 @@ mongodb {
   uri = "mongodb://localhost:27017/help-to-save"
 }
 
+# Increase default platform timeout from 20s to (FE)45s (BE)43s which is more than DES at 30s
+play.ws.timeout.request = 43s
```
sureshhmrc: are we interested in `play.ws.timeout.connection` as well ?

PR #75, opened 2018-01-08
Mask NINO before logging messages

sureshhmrc:


app/uk/gov/hmrc/helptosave/util/HttpResponseOps.scala
```diff
@@ -35,15 +35,15 @@ class HttpResponseOps(val response: HttpResponse) extends AnyVal {
     Try(response.json).fold(
       error ⇒
         // response.json failed in this case - there was no JSON in the response
-        Left(s"Could not read http response as JSON (${error.getMessage}). Response body was ${response.body}"),
```
AndyHWChung: are there cases where we do want to log the response body? Wouldn't this work?
```scala
 Response body was ${maskNino(response.body)}
```

sureshhmrc: there is no harm is doing this change - let me fix it

AndyHWChung:
LGTM

PR #81, opened 2018-01-18
HTS-676: Add "/create-de-account" endpoint

sureshhmrc:


app/uk/gov/hmrc/helptosave/controllers/CreateDEAccountController.scala
```diff
@@ -0,0 +1,52 @@
...
+            frontendConnector.createAccount(userInfo)
+              .map(response ⇒
+                Option(response.body).fold[Result](
+                  Status(response.status)
+                )(body ⇒ Status(response.status)(body)))
+
+          case Some(error: JsError) ⇒
+            val errorString = error.prettyPrint()
+            logger.warn(s"Could not parse JSON in request body: $errorString")
+            toFuture(BadRequest(ErrorResponse("Could not parse JSON in request", errorString).toJson()))
```
AndyHWChung: do you need to call `toFuture` explicitly here? Will the implicit conversion happen not happen without it?

app/uk/gov/hmrc/helptosave/connectors/FrontendConnector.scala
```diff
@@ -0,0 +1,43 @@
...
+}
+
+@Singleton
+class FrontendConnectorImpl @Inject() (http: WSHttp)(implicit hc: HeaderCarrier)
+  extends FrontendConnector with ServicesConfig with Logging {
+
+  val createAccountURL: String = getString("microservice.services.help-to-save-frontend.url")
+
+  override def createAccount(userInfo: NSIUserInfo)(implicit hc: HeaderCarrier, ec: ExecutionContext): Future[HttpResponse] = {
+    http.post(createAccountURL, userInfo)
```
AndyHWChung: do we want to do a `recover` block here to better control the error response in case the future fails?

app/uk/gov/hmrc/helptosave/controllers/CreateDEAccountController.scala
```diff
@@ -0,0 +1,52 @@
...
+import uk.gov.hmrc.helptosave.models.{ErrorResponse, NSIUserInfo}
+import uk.gov.hmrc.helptosave.util.JsErrorOps._
+import uk.gov.hmrc.helptosave.util.{Logging, toFuture}
+import uk.gov.hmrc.play.microservice.controller.BaseController
+
+class CreateDEAccountController @Inject() (frontendConnector: FrontendConnector) extends BaseController with Logging with WithMdcExecutionContext {
+
+  def createDEAccount(): Action[AnyContent] = Action.async {
+    implicit request ⇒
+      {
```
AndyHWChung: this is minor - it looks like you don't need these braces here

AndyHWChung:
LGTM

PR #85, opened 2018-02-09
HTS-764: Pass UC params to DES eligibility check

sureshhmrc:


app/uk/gov/hmrc/helptosave/models/UCResponse.scala
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.helptosave.models
+
+import play.api.libs.json.{Format, Json}
+
+case class UCResponse(ucClaimant: String, withinThreshold: Option[String])
```
AndyHWChung: it is more meaningful to have these parameters as Booleans and write a custom reads and writes instance that reads and writes it to "Y" and "N" - this would provide a bit of validation on the response from DWP as well which is missing at the moment

app/uk/gov/hmrc/helptosave/models/UCResponse.scala
```diff
@@ -16,10 +16,32 @@
...
-case class UCResponse(ucClaimant: String, withinThreshold: Option[String])
+case class UCResponse(ucClaimant: Boolean, withinThreshold: Boolean)
 
 object UCResponse {
-  implicit val format: Format[UCResponse] = Json.format[UCResponse]
+
+  implicit val reads: Format[UCResponse] = new Format[UCResponse] {
+
+    override def reads(json: JsValue): JsResult[UCResponse] = {
+      ((json \ "ucClaimant").as[String], (json \ "withinThreshold").asOpt[String]) match {
```
AndyHWChung: `.as` will throw if the reading of the field to `String` is unsuccessful. Using `.validate` and `.validateOpt` will allow for better error handling.

app/uk/gov/hmrc/helptosave/models/UCResponse.scala
```diff
@@ -16,10 +16,32 @@
 
 package uk.gov.hmrc.helptosave.models
 
-import play.api.libs.json.{Format, Json}
+import play.api.libs.json._
 
-case class UCResponse(ucClaimant: String, withinThreshold: Option[String])
+case class UCResponse(ucClaimant: Boolean, withinThreshold: Boolean)
```
AndyHWChung: `withinThreshold` should be an `Option[Boolean]` - if it is not there we don't want to send that parameter to DES

app/uk/gov/hmrc/helptosave/models/UCResponse.scala
```diff
@@ -0,0 +1,58 @@
...
+          errors ⇒ JsError(s"unable to parse UCResponse from proxy, due to=$errors"),
+          a ⇒
+            (json \ "withinThreshold").validateOpt[String]
+              .fold(
+                errors ⇒ JsError(s"unable to parse UCResponse from proxy, due to=$errors"),
+                b ⇒
+                  (a, b) match {
+                    case ("Y", Some("Y")) ⇒ JsSuccess(UCResponse(ucClaimant      = true, withinThreshold = Some(true)))
+                    case ("Y", Some("N")) ⇒ JsSuccess(UCResponse(ucClaimant      = true, withinThreshold = Some(false)))
+                    case ("N", _)         ⇒ JsSuccess(UCResponse(ucClaimant      = false, withinThreshold = Some(false)))
```
AndyHWChung: in this case `withinThreshold` would be `None`

app/uk/gov/hmrc/helptosave/models/UCResponse.scala
```diff
@@ -0,0 +1,58 @@
...
+              )
+        )
+    }
+
+    override def writes(response: UCResponse): JsValue = {
+
+      val (a, b) = response match {
+        case UCResponse(true, Some(true))  ⇒ ("Y", "Y")
+        case UCResponse(true, Some(false)) ⇒ ("Y", "N")
+        case _                             ⇒ ("N", "N")
```
AndyHWChung: we shouldn't have a second field in this case - this would be mapped to `None`

app/uk/gov/hmrc/helptosave/connectors/EligibilityCheckConnector.scala
```diff
@@ -47,17 +47,21 @@ class EligibilityCheckConnectorImpl @Inject() (http:              WSHttp,
 
   val itmpBaseURL: String = baseUrl("itmp-eligibility-check")
 
-  def url(nino: String): String =
-    s"$itmpBaseURL/help-to-save/eligibility-check/$nino"
+  def url(nino: String, ucResponse: Option[UCResponse]): String = {
+    ucResponse match {
+      case Some(UCResponse(a, b)) ⇒ s"$itmpBaseURL/help-to-save/eligibility-check/$nino?ucClaimant=$a&withinThreshold=${b.getOrElse(false)}"
```
AndyHWChung: if `withinThreshold` is `None` we don't want to send that parameter to DES. We'd do something like this:
```scala
s"$itmpBaseURL/help-to-save/eligibility-check/$nino?ucClaimant=$a${b.map(x => s"&withinThreshold=$x").getOrElse("")}"
```

AndyHWChung: we basically want to pass in the parameters as DWP give them to us where the `withinThreshold` field is optional

sureshhmrc: ah yes, i missed the important bit -  fixing it now..

AndyHWChung:
LGTM

PR #86, opened 2018-02-15
change logging from info to debug to reduce noise in the logs

sureshhmrc:


sureshhmrc:
https://kibana.tools.production.tax.service.gov.uk/goto/a60c5ad160aecac659daebf01a858ef3

AndyHWChung:
You're right - those logs are very noisy. If we want to see how long those calls took we've can look in grafana for that since we have custom metrics for those times.

PR #89, opened 2018-02-19
HTS-766: Moved Auditing for Eligibility checks from frontend

sureshhmrc:


app/uk/gov/hmrc/helptosave/models/HTSEvent.scala
```diff
@@ -0,0 +1,77 @@
...
+        val reason = "Response: " +
+          s"resultCode=${eligibilityResult.resultCode}, reasonCode=${eligibilityResult.reasonCode}, " +
+          s"meaning result='${eligibilityResult.result}', reason='${eligibilityResult.reason}'"
+
+        ucResponse match {
+          case Some(ucR) ⇒ Map[String, String](
+            "nino" → nino,
+            "eligible" → "false",
+            "isUCClaimant" → ucR.ucClaimant.toString,
+            "isWithinUCThreshold" → ucR.withinThreshold.getOrElse(false).toString,
```
AndyHWChung: it'd be more accurate to not include the threshold field in the event when there isn't one there

AndyHWChung:
or:
```scala
def ucData(ucResponse: Option[UCResponse]): Map[String,String] = ucResponse match {
  case Some(UCResponse(isClaimant, Some(withinThreshold))) => Map(...)
  case Some(UCResponse(isClaimant, None)) => Map(...)
  case None => Map.empty[String,String]
 }
```
might be easier to understand

sureshhmrc:
fixed !

AndyHWChung:
LGTM

PR #94, opened 2018-02-23
HTS-847: Log "X-CorrelationId" and pass it to downstream services

sureshhmrc:


app/uk/gov/hmrc/helptosave/util/Logging.scala
```diff
@@ -29,42 +32,52 @@ object Logging {
...
     if (configuration.underlying.getBoolean("nino-logging.enabled")) {
       nino ⇒ s"For NINO [$nino]: "
     } else {
       _ ⇒ ""
     }
 
-  def transform(message: String, nino: NINO): String = loggingPrefix(nino) + message
+  private val correlationIdPrefix: Option[String] ⇒ String =
+    {
+      case Some(id) ⇒ s", For CorrelationId $id"
```
AndyHWChung: We would want some sort of separator and a space here to make the log messages legible. So if the correlation id was `abc` and there was no NINO we don't want something like this in the logs:
```
", For CorrelationId abcSome log message " 
```

app/uk/gov/hmrc/helptosave/util/Logging.scala
```diff
@@ -29,42 +32,52 @@ object Logging {
...
-      logger.error(transformer.transform(message, nino))
+    def error(message: String, nino: NINO)(implicit transformer: LogMessageTransformer, hc: HeaderCarrier): Unit =
+      logger.error(transformer.transform(message, nino, getCorrelationId()))
 
-    def error(message: String, e: ⇒ Throwable, nino: NINO)(implicit transformer: NINOLogMessageTransformer): Unit =
-      logger.error(transformer.transform(message, nino), e)
+    def error(message: String, e: ⇒ Throwable, nino: NINO)(implicit transformer: LogMessageTransformer, hc: HeaderCarrier): Unit =
+      logger.error(transformer.transform(message, nino, getCorrelationId()), e)
 
+    def getCorrelationId()(implicit hc: HeaderCarrier): Option[String] =
```
AndyHWChung: `getCorrelationId` doesn't belong in the implicit class for `Logger` - it doesn't extend the behaviour of `Logger`, it extends the behaviour of `HeaderCarrier`. It would make more sense have a private method outside of this implicit class to get the behaviour you want

app/uk/gov/hmrc/helptosave/util/Logging.scala
```diff
@@ -29,42 +32,52 @@ object Logging {
...
+      hc.headers.find(p ⇒ p._1 === "X-CorrelationId").map(_._2)
```
AndyHWChung: The header name is configurable in the API repo - reading the header key from config here would make it more flexible

app/uk/gov/hmrc/helptosave/util/Logging.scala
```diff
@@ -29,42 +32,52 @@ object Logging {
 
   implicit class LoggerOps(val logger: Logger) {
 
-    def debug(message: String, nino: NINO)(implicit transformer: NINOLogMessageTransformer): Unit =
-      logger.debug(transformer.transform(message, nino))
+    def debug(message: String, nino: NINO)(implicit transformer: LogMessageTransformer, hc: HeaderCarrier): Unit =
```
AndyHWChung: where we're using a `Logger` we may not always have an instance of `HeaderCarrier` - making a `HeaderCarrier` an implicit parameter here restricts the use of the methods in `LoggerOps`. I think it'd be more flexible if we created extra methods in `LoggerOps` to pass in `correlationId: Option[UUID]` with perhaps an default value of `None` if we don't want to create extra noise

AndyHWChung: The same goes for the proxy PR as well

AndyHWChung:
LGTM

PR #101, opened 2018-04-06
HTS-928: Add countryCode and PhoneNumber to the paye-personal-details endpoint

sureshhmrc:


app/uk/gov/hmrc/helptosave/models/CountryCode.scala
```diff
@@ -0,0 +1,40 @@
...
+import play.api.libs.json._
+
+import scala.collection.Map
+import scala.io.Source
+
+case class CountryCode(code: Option[String])
+
+object CountryCode {
+
+  private val countryCodes: Map[String, Option[String]] = {
```
sureshhmrc: so map.get("1") returns something like `Some(CountryCode(Some("GB")))` - trying to simplify the logic - but ideas welcome

app/uk/gov/hmrc/helptosave/models/PayePersonalDetails.scala
```diff
@@ -85,15 +84,61 @@ object PayePersonalDetails {
...
+                  JsSuccess(None)
+                ) { v ⇒
+                    val callingCode = (v \ "callingCode").asOpt[Int]
+                    val convertedAreaDiallingCode = (v \ "convertedAreaDiallingCode").asOpt[String]
+                    val telephoneNumber = (v \ "telephoneNumber").asOpt[String]
+
+                    (callingCode, convertedAreaDiallingCode, telephoneNumber) match {
+                      case (Some(ccKey), Some(cadc), Some(t)) ⇒
+                        val cc = CallingCodes.getCodeFor(ccKey).getOrElse("")
+                        JsSuccess(Some(s"+$cc${cadc.replaceFirst("^0+(?!$)", "")}$t"))
```
sureshhmrc: `replaceFirst` replaces all the leading `zeros` in the string - currently this fails for `250 - 0` which has only one `0` and it will be replaced by `""`

app/uk/gov/hmrc/helptosave/models/CallingCodes.scala
```diff
@@ -0,0 +1,42 @@
...
+
+import scala.collection.Map
+import scala.io.Source
+
+object CallingCodes {
+
+  private val callingCodes: Map[Int, String] = {
+    var codes = Map[Int, String]()
+    val content = Source.fromInputStream(getClass.getResourceAsStream("/resources/callingcodes.txt")).getLines()
+    content.foreach {
```
AndyHWChung: you could do a `foldLeft` here instead of having a mutable `var` and updating it in a loop. Then you could have something like this:
```scala
content.foldLeft(Map.empty[Int,String]){ case (acc, curr) =>
  curr.split("-").toList match {
    case key :: value :: Nil => acc.updated(key.toInt, value)
    case _ => acc
  }
}
```
Note in the above I've avoided doing a `get` on the array and got the key-value pair via pattern matching

app/uk/gov/hmrc/helptosave/models/CountryCode.scala
AndyHWChung: it would be more meaningful to have the type as `Map[String,String]` and not populate the map with values that don't have a value. Also, doesn't the countryCode key come as an `Int` so that really we should have a ``Map[Int,String]`?

app/uk/gov/hmrc/helptosave/models/PayePersonalDetails.scala
AndyHWChung: I think you could just do `cadc.stripPrefix("0")` here

app/uk/gov/hmrc/helptosave/models/CountryCode.scala
```diff
@@ -0,0 +1,40 @@
...
+ */
+
+package uk.gov.hmrc.helptosave.models
+
+import play.api.libs.json._
+
+import scala.collection.Map
+import scala.io.Source
+
+case class CountryCode(code: Option[String])
```
AndyHWChung: is this case class actually used anywhere?

app/uk/gov/hmrc/helptosave/models/CountryCode.scala
```diff
@@ -0,0 +1,36 @@
...
+object CountryCode {
+
+  val countryCodes: Map[Int, String] = {
+    val content = Source.fromInputStream(getClass.getResourceAsStream("/resources/country.json")).mkString
+    Json.parse(content) match {
+      case JsObject(fields) ⇒
+        fields
+          .map(x ⇒ (x._1.toInt, (x._2 \ "alpha_two_code").asOpt[String]))
+          .collect { case (id, Some(value)) ⇒ id -> value }
+      case _ ⇒ Map.empty[Int, String]
```
AndyHWChung: in this case we probably want to be informed that no country codes were found - the application shouldn't be allowed to start if the country code file was sable to be parsed properly

app/uk/gov/hmrc/helptosave/models/PayePersonalDetails.scala
```diff
@@ -85,15 +86,61 @@ object PayePersonalDetails {
         def readDob(): JsResult[LocalDate] =
           readData[LocalDate](List(read("dateOfBirth")), "DateOfBirth")
 
-        def readAddress(): JsResult[Address] =
-          readData[Address](List(readSeq("addresses", "2"), readSeq("addresses", "1")), "Address") // 1–Residential Address, 2–Correspondence Address
+        def readAddress(): JsResult[Address] = {
+          List(readSeq("addresses", "2"), readSeq("addresses", "1")) //7–Mobile Telephone Number, 1–Daytime Home Telephone Number
```
AndyHWChung: the comment here is correct for phone numbers rather than addresses

app/uk/gov/hmrc/helptosave/models/PayePersonalDetails.scala
```diff
@@ -85,15 +86,61 @@ object PayePersonalDetails {
...
+                x.fold[JsResult[Option[String]]](
+                  JsSuccess(None)
+                ) { v ⇒
+                    val callingCode = (v \ "callingCode").asOpt[Int]
+                    val convertedAreaDiallingCode = (v \ "convertedAreaDiallingCode").asOpt[String]
+                    val telephoneNumber = (v \ "telephoneNumber").asOpt[String]
+
+                    (callingCode, convertedAreaDiallingCode, telephoneNumber) match {
+                      case (Some(ccKey), Some(cadc), Some(t)) ⇒
+                        val cc = callingCodes.getOrElse(ccKey, "")
```
AndyHWChung: if the country calling code is blank, putting a `+` and stripping out the leading zero of the area code would lead to an incorrect numbere. Perhaps what is needed is something like:
```scala
(callingCode.flatMap(callingCodes.get), convertedAreaDiallingCode, telephoneNumber) match {
 ...
}
```

AndyHWChung:
LGTM

PR #127, opened 2018-05-10
HTS-943: Added UCThreshold Startup Behaviour

Jackie1988:


app/uk/gov/hmrc/helptosave/actors/ThresholdConnectorProxy.scala
```diff
@@ -0,0 +1,48 @@
...
+
+import akka.pattern.pipe
+import akka.actor.{Actor, Props}
+import uk.gov.hmrc.helptosave.actors.ThresholdConnectorProxy.{GetThresholdValue, GetThresholdValueResponse}
+import uk.gov.hmrc.helptosave.connectors.ThresholdConnector
+import uk.gov.hmrc.http.HeaderCarrier
+
+import scala.concurrent.{ExecutionContext, Future}
+
+class ThresholdConnectorProxy(thresholdConnector: ThresholdConnector) extends Actor {
```
sureshhmrc: can we rename this to `UCThresholdActor` ? or something like that

app/uk/gov/hmrc/helptosave/actors/ThresholdMongoProxy.scala
```diff
@@ -0,0 +1,55 @@
...
+
+package uk.gov.hmrc.helptosave.actors
+
+import akka.pattern.pipe
+import akka.actor.{Actor, Props}
+import cats.syntax.either._
+import uk.gov.hmrc.helptosave.actors.ThresholdMongoProxy._
+import uk.gov.hmrc.helptosave.repo.ThresholdStore
+
+class ThresholdMongoProxy(thresholdStore: ThresholdStore) extends Actor {
```
sureshhmrc: @AndyHWChung why do we need to do this ? 
we just store threshold in the mongo store and thats it - we don't care even if there are issues during storing to mongo because the next call after 24h will make another update ?

app/uk/gov/hmrc/helptosave/actors/UCThresholdManager.scala
```diff
@@ -0,0 +1,192 @@
...
+
+    case _: StoreMongoThresholdValue ⇒
+      // write value to mongo
+      logger.info("Trying to store threshold value in mongo")
+      storeValue(thresholdValue) pipeTo self
+
+    case StoreMongoThresholdValueResponse(result) ⇒
+      result.fold({
+        e ⇒
+          storeMongoRetries.retry(thresholdValue).fold(
```
sureshhmrc: @AndyHWChung  i would say we are overkilling here - we dont need to retry storing in mongo - even if the call fails we just ignore it as we will anyway do another update after 24h

app/uk/gov/hmrc/helptosave/actors/UCThresholdManager.scala
```diff
@@ -0,0 +1,192 @@
...
+          storeMongoRetries.cancelAndReset()
+      })
+
+  }
+
+  def handleDESThresholdValue(result: Either[String, Double]): Unit = result.fold(
+    { e ⇒
+      logger.warn(s"Could not obtain the threshold value from DES: $e")
+      //If Des is down, try mongo
+      self ! GetMongoThresholdValue
```
sureshhmrc: over kiil again ?

app/uk/gov/hmrc/helptosave/actors/UCThresholdManager.scala
```diff
@@ -0,0 +1,192 @@
...
+        context become ready(value)
+        self ! StoreMongoThresholdValue(value)
+        unstashAll()
+    }
+  )
+
+  def handleMongoThresholdValue(result: Either[String, Option[Double]]): Unit = {
+      def handleError(): Unit = {
+        getDESRetries.retry(())
+        pagerDutyAlerting.alert("Could not initialise threshold value")
```
sureshhmrc: i would say just raise a PDAlert but dont do any retries

app/uk/gov/hmrc/helptosave/connectors/ThresholdConnector.scala
```diff
@@ -0,0 +1,80 @@
...
+import uk.gov.hmrc.helptosave.models.UCThreshold
+import uk.gov.hmrc.helptosave.util.HttpResponseOps._
+import uk.gov.hmrc.helptosave.util.{LogMessageTransformer, Logging, PagerDutyAlerting, Result}
+import uk.gov.hmrc.helptosave.util.Logging._
+import uk.gov.hmrc.http.HeaderCarrier
+
+import scala.concurrent.{ExecutionContext, Future}
+
+@ImplementedBy(classOf[ThresholdConnectorImpl])
+trait ThresholdConnector {
```
sureshhmrc: `UCThresholdConnector` ?

app/uk/gov/hmrc/helptosave/connectors/ThresholdConnector.scala
```diff
@@ -0,0 +1,80 @@
...
+
+  def getThreshold()(implicit hc: HeaderCarrier, ec: ExecutionContext): Result[Double]
```
sureshhmrc: `getUCThreshold` ?

app/uk/gov/hmrc/helptosave/connectors/ThresholdConnector.scala
```diff
@@ -0,0 +1,80 @@
...
+
+              case other ⇒
+                logger.warn(s"Call to get threshold unsuccessful. Received unexpected status $other. " +
+                  s"Body was: ${response.body}", "-", additionalParams)
+                pagerDutyAlerting.alert("Received unexpected http status in response to get threshold")
+                Left(s"Received unexpected status $other")
+            }
+          }.recover {
+            case e ⇒
+              pagerDutyAlerting.alert("Failed to make call to get threshold")
```
sureshhmrc: dont forget to add these PD messages to `alert-config` service

app/uk/gov/hmrc/helptosave/connectors/ThresholdConnector.scala
```diff
@@ -0,0 +1,80 @@
...
+              Left(s"Call to get threshold unsuccessful: ${e.getMessage}")
```
sureshhmrc: `call to get UCThreshold unsuccessful` ?

app/uk/gov/hmrc/helptosave/util/WithExponentialBackoffRetry.scala
```diff
@@ -0,0 +1,133 @@
...
+
+package uk.gov.hmrc.helptosave.util
+
+import akka.actor.{Actor, ActorRef, Cancellable, Scheduler}
+import uk.gov.hmrc.helptosave.util.WithExponentialBackoffRetry.ExponentialBackoffRetry
+
+import scala.concurrent.ExecutionContext
+import scala.concurrent.duration._
+
+trait WithExponentialBackoffRetry { this: Actor ⇒
```
sureshhmrc: @AndyHWChung  again, unless i miss some requirement , this is an over kill

conf/application.conf
```diff
@@ -48,6 +48,9 @@ play.modules.enabled += "uk.gov.hmrc.play.bootstrap.HttpClientModule"
 # Provides an implementation and configures all filters required by a Platform microservice.
 play.modules.enabled += "uk.gov.hmrc.play.bootstrap.MicroserviceModule"
 
+# Starts up the UC THreshold Orchestrator
+play.modules.enabled += "uk.gov.hmrc.helptosave.modules.UCThresholdModule"
```
sureshhmrc: do we have this class `UCThresholdModule ` anywhere I dont see it in the PR ?

conf/application.conf
```diff
@@ -48,6 +48,9 @@ play.modules.enabled += "uk.gov.hmrc.play.bootstrap.HttpClientModule"
 # Provides an implementation and configures all filters required by a Platform microservice.
 play.modules.enabled += "uk.gov.hmrc.play.bootstrap.MicroserviceModule"
 
+# Starts up the UC THreshold Orchestrator
```
sureshhmrc: `THreshold ` -> `Threshold `

app/uk/gov/hmrc/helptosave/actors/ThresholdMongoProxy.scala
AndyHWChung: When the backend starts we won't have an in-memory value to use and in this state no eligibility checks can be done. We need to retrieve a value as quickly as possible and if DES is down we use mongo as a backup. In this scenario, we need to make sure that the value in mongo is as up-to-date as possible.
The chances that mongo has the wrong value are very slim: it'd have to be the day when the change is made and the write to mongo failed and then we redeployed the application before the updated value was stored to mongo. But since this chance still exists, it is safer to ensure we definitely write the value to mongo.

app/uk/gov/hmrc/helptosave/connectors/ThresholdConnector.scala
AndyHWChung: yes, we need to remember

app/uk/gov/hmrc/helptosave/util/WithExponentialBackoffRetry.scala
AndyHWChung: when we retry we don't want to spam the service we're trying to reach - if they're down then a large number of requests to them isn't going to help to situation. see here for some info: https://dzone.com/articles/understanding-retry-pattern-with-exponential-back

conf/application.conf
AndyHWChung: you're right, I can't see it either. We'll investigate...

app/uk/gov/hmrc/helptosave/actors/ThresholdConnectorProxy.scala
AndyHWChung: the actor is really acting as a proxy to a threshold connector so the name given is more descriptive then UCThresholdActor I think

app/uk/gov/hmrc/helptosave/actors/ThresholdMongoProxy.scala
AndyHWChung: in reality, we're not expecting that writes to mongo are going to fail for a long period of time - but at least if it is we're prepared

conf/application.conf
AndyHWChung: we've pushed a new commit with the missing stuff included. Turns out gitignore was ignoring the files

app/uk/gov/hmrc/helptosave/actors/ThresholdConnectorProxy.scala
sureshhmrc: Generally actor classes names end with *Actor - so thought of renaming this to ***Actor - and i see its not just acting as a proxy to `thresholdConnector `  because its doing actor stuff in it

AndyHWChung: I guess `ThresholdConnectorProxyActor` is more descriptive then?

app/uk/gov/hmrc/helptosave/actors/ThresholdMongoProxy.scala
AndyHWChung: if the write to mongo fails and we don't retry there is still the danger that the value read from mongo on the next application startup is going to be wrong. The retries are only done when there is an error writing to mongo - the retries are stopped once a successful write occurs

app/uk/gov/hmrc/helptosave/util/WithExponentialBackoffRetry.scala
AndyHWChung: in this PR, there is no daily scheduled retry at the moment - that will be built in later. The retries in this PR are retries for when the application starts up and there are error retriving a threshold value and errors writing to mongo

app/uk/gov/hmrc/helptosave/actors/ThresholdConnectorProxy.scala
sureshhmrc: hmm, thats fine, i will leave it up to @Jackie1988  :)

app/uk/gov/hmrc/helptosave/actors/ThresholdMongoProxy.scala
sureshhmrc: `the value read from mongo on the next application startup is going to be wrong` -  yes, this is actually an issue only if the update that failed tried to update with a new/updated threshold value which is a rare case  given that how frequently the ITMP threshold value changes...
any way there is no harm in keeping the re-try logic in the codebase...

app/uk/gov/hmrc/helptosave/actors/UCThresholdManager.scala
sureshhmrc: can you make the message a bit more clear ?

sureshhmrc:
I still don't see the `UCThresholdModule ` being pushed ?

app/uk/gov/hmrc/helptosave/actors/ThresholdMongoProxy.scala
```diff
@@ -0,0 +1,55 @@
...
+  import context.dispatcher
+
+  def retrieveThreshold() = thresholdStore.getThreshold().value
```
sureshhmrc: can we deal with the case class `UCThreshold` every where here which wraps the double - not sure if it makes things any better ?

app/uk/gov/hmrc/helptosave/connectors/ThresholdConnector.scala
Jackie1988: I have added them in a new branch

app/uk/gov/hmrc/helptosave/util/WithExponentialBackoffRetry.scala
AndyHWChung: it would read the value returned by DES in that case - we only try to write to mongo when we've got a value from DES

app/uk/gov/hmrc/helptosave/actors/ThresholdMongoProxy.scala
AndyHWChung: the case class is being used to read the JSON response from the connector and to write to mongo in the ThresholdStore class. In between though, it would be simpler to pass in doubles if this is not the case

AndyHWChung: I think we are mostly doing that though

AndyHWChung:
the module stuff should be there now - we pushed another commit

app/uk/gov/hmrc/helptosave/actors/ThresholdMongoProxy.scala
sureshhmrc: then its no point to use the class `UCThreshold` here in the mongo store
```  private[repo] def doUpdate(amount: Double)(implicit ec: ExecutionContext): Future[Option[UCThreshold]] =```

**Comment by sureshhmrc** [comment]
code looks good to me

File: app/uk/gov/hmrc/helptosave/repo/UCThresholdStore.scala [hunk 14]
```diff
@@ -0,0 +1,85 @@
...
+
+  def storeUCThreshold(amount: Double)(implicit ec: ExecutionContext): EitherT[Future, String, Unit]
+
+  def getUCThreshold()(implicit ec: ExecutionContext): EitherT[Future, String, Option[Double]]
+}
+
+@Singleton
+class MongoThresholdStore @Inject() (mongo: ReactiveMongoComponent)(implicit transformer: LogMessageTransformer)
+  extends ReactiveRepository[UCThreshold, BSONObjectID](
+    collectionName = "threshold",
```
sureshhmrc: @Jackie1988  can you rename this to `UCThreshold` something like that to make the collection more specific ?
and also remove unused imports in `UCThresholdMongoProxySpec` ? 
thanks

Jackie1988: ok sure, done, thanks

PR #133, opened 2018-05-23
HTS-1075: Modified Threshold Retrieval from DES

Jackie1988:


app/uk/gov/hmrc/helptosave/actors/UCThresholdManager.scala
```diff
@@ -17,26 +17,22 @@
...
 import akka.pattern.ask
 import akka.util.Timeout
 import com.typesafe.config.Config
-import uk.gov.hmrc.helptosave.util.{Logging, PagerDutyAlerting, WithExponentialBackoffRetry}
+import uk.gov.hmrc.helptosave.util.{Logging, PagerDutyAlerting, Time, WithExponentialBackoffRetry}
 
 import scala.concurrent.duration._
 import scala.concurrent.Future
-
 import UCThresholdConnectorProxy.{GetThresholdValue ⇒ GetDESThresholdValue, GetThresholdValueResponse ⇒ GetDESThresholdValueResponse}
```
sureshhmrc: do we still need these shady names in this import ?

app/uk/gov/hmrc/helptosave/actors/UCThresholdManager.scala
```diff
@@ -88,79 +67,31 @@ class UCThresholdManager(thresholdConnectorProxy: ActorRef,
...
-
-    case GetMongoThresholdValue ⇒
-      logger.info("Trying to get UC threshold value from mongo")
-      getValueFromMongo() pipeTo self
-
-    case GetMongoThresholdValueResponse(result) ⇒
-      handleMongoThresholdValue(result)
+      result.fold(
+        { e ⇒
+          pagerDutyAlerting.alert("Could not obtain initial UC threshold value from DES")
```
sureshhmrc: do we need to do this here ?  or inside of `fold()` below ? because if des is down and if we retry several times with in a minute, then it causes several alerts  (number of alerts actually depend on how frequent the retries are, anyway)

sureshhmrc: Also can you rename `UCThresholdConnectorProxy` to something that ends with ***Actor  due to this class not being a `Proxy` and its placed inside `actor` package ?

app/uk/gov/hmrc/helptosave/actors/UCThresholdManager.scala
```diff
@@ -185,9 +115,8 @@ object UCThresholdManager {
   case class GetThresholdValueResponse(result: Double)
 
   def props(thresholdConnectorProxy: ActorRef,
-            thresholdMongoProxy:     ActorRef,
             pagerDutyAlerting:       PagerDutyAlerting,
             scheduler:               Scheduler,
             config:                  Config): Props =
-    Props(new UCThresholdManager(thresholdConnectorProxy, thresholdMongoProxy, pagerDutyAlerting, scheduler, config))
+    Props(new UCThresholdManager(thresholdConnectorProxy, pagerDutyAlerting, scheduler, config))
```
sureshhmrc: are you making use of `Command` in the above line `sealed trait Command` anywhere ?

app/uk/gov/hmrc/helptosave/actors/UCThresholdManager.scala
```diff
@@ -88,79 +67,31 @@ class UCThresholdManager(thresholdConnectorProxy: ActorRef,
       getValueFromDES() pipeTo self
 
     case GetDESThresholdValueResponse(result) ⇒
```
sureshhmrc: do we need to have both cases `GetDESThresholdValue` and `GetDESThresholdValueResponse` cant we merge both into one case ? 
also whats the benefit of `context become ready(value)` line at the moment ?

app/uk/gov/hmrc/helptosave/metrics/Metrics.scala
```diff
@@ -65,54 +63,3 @@ class Metrics @Inject() (metrics: com.kenshoo.play.metrics.Metrics) {
   val getAccountErrorCounter: Counter = counter("backend.get-account-error.count")
 }
 
-object Metrics {
```
sureshhmrc: unused tailrec import in this class ?

app/uk/gov/hmrc/helptosave/actors/UCThresholdManager.scala
Jackie1988: Yeah we need to keep this to differentiate between case classes in the UCThresholdManager object and case classes in the UCThresholdConnectorProxy object

app/uk/gov/hmrc/helptosave/actors/UCThresholdManager.scala
Jackie1988: We won't cause several pd alerts due to how its set up in the background. I have renamed UCThresholdConnectorProxy to UCThresholdConnectorProxyActor

sureshhmrc: the class is not actually a `proxy` -  but if you have some reason to use `proxy` thats fine

app/uk/gov/hmrc/helptosave/metrics/Metrics.scala
Jackie1988: Removed, thanks

app/uk/gov/hmrc/helptosave/actors/UCThresholdManager.scala
Jackie1988: No so I have removed it, thank you

app/uk/gov/hmrc/helptosave/actors/UCThresholdManager.scala
Jackie1988: I'm leaving the name as it is

PR #134, opened 2018-05-30
HTS-1048 - refactored get account logic

AndyHWChung:


app/uk/gov/hmrc/helptosave/controllers/AccountController.scala
```diff
@@ -29,11 +33,23 @@ class AccountController @Inject() (proxyConnector: HelpToSaveProxyConnector,
...
+      Forbidden
+    } else {
+      val id = correlationId.getOrElse(UUID.randomUUID().toString)
+      proxyConnector.getAccount(nino, systemId, id)
+        .fold(
+          { e ⇒
+            logger.warn(e)
+            InternalServerError
+          },
+          _.fold[Result](NotFound)(account ⇒ Ok(Json.toJson(account)))
```
sureshhmrc: if there are any validation errors (but the account still exists) then it returns `NotFound ` to the caller , does this make sense  ?

app/uk/gov/hmrc/helptosave/controllers/HelpToSaveAuth.scala
```diff
@@ -46,10 +46,10 @@ class HelpToSaveAuth(htsAuthConnector: AuthConnector) extends BaseController wit
     Action.async { implicit request ⇒
       authorised(AuthWithCL200)
         .retrieve(Retrievals.nino) { mayBeNino ⇒
-          mayBeNino.fold(
-            toFuture(InternalServerError("could not find NINO for logged in user"))
+          mayBeNino.fold[Future[Result]](
+            Forbidden
```
sureshhmrc: this case never happens ideally , but can we add a log line here saying something like "nino not found for a logged in user" ?

app/uk/gov/hmrc/helptosave/controllers/HelpToSaveAuth.scala
```diff
@@ -58,15 +58,15 @@ class HelpToSaveAuth(htsAuthConnector: AuthConnector) extends BaseController wit
   def handleFailure(): PartialFunction[Throwable, Result] = {
     case _: NoActiveSession ⇒
       logger.warn("user is not logged in, probably a hack?")
-      Forbidden("no active session found for logged in user")
+      Unauthorized
 
-    case _: InsufficientConfidenceLevel | _: InsufficientEnrolments ⇒
-      logger.warn("unexpected: not met required ConfidenceLevel for logged in user")
-      Forbidden("not met required ConfidenceLevel for logged in user")
+    case e: InternalError ⇒
```
sureshhmrc: what happens in this case ? `_: InsufficientConfidenceLevel | _: InsufficientEnrolments `

app/uk/gov/hmrc/helptosave/controllers/AccountController.scala
AndyHWChung: validation errors get mapped to `InternalServerError` - the `None` case here is when NS&I have said there is no account

app/uk/gov/hmrc/helptosave/controllers/HelpToSaveAuth.scala
AndyHWChung: yes, I'll add that back in

app/uk/gov/hmrc/helptosave/controllers/HelpToSaveAuth.scala
AndyHWChung: those will get picked up in the `AuthorisationException` case

app/uk/gov/hmrc/helptosave/controllers/AccountController.scala
sureshhmrc: ok cool, 
but how does the caller distinguish between a real ISE and validation errors,
if there are any validation errors, why don't we return something meaningful to the caller and then caller , based on the repose, says "your account is in inconsistent state, please call us...blah...", 
because if user sees ISE, then he might be keep trying to close his account everyday and he says same ISE :) :)

test/uk/gov/hmrc/helptosave/connectors/HelpToSaveProxyConnectorSpec.scala
```diff
@@ -293,17 +295,46 @@ class HelpToSaveProxyConnectorSpec extends TestSupport with MockPagerDuty with E
...
+              |      "errorMessageId" : "${appConfig.runModeConfiguration.underlying.getString("nsi.get-account.no-account-error-message-id")}",
+              |      "errorMessage"   : "Oh no!",
+              |      "errorDetail"    : "Account doesn't exist"
+              |    }
+              |  ]
+              |}
+          """.stripMargin)
+
+        mockGetAccountResponse(getAccountUrl)(Some(HttpResponse(400, Some(errorResponse))))
+        mockPagerDutyAlert("Received unexpected http status in response to getAccount")
```
frankoid: I'm not sure whether you'd want an alert in this case, it could be normal for the specified NINO to not have a Help to Save account (maybe a caller might use this endpoint to find out whether an account exists for the specified NINO).

In the short term, given the way the mobile services call it I wouldn't expect this alert to be fired so you could leave it in until/unless you notice a problem with false alarms.

Another way of dealing with it could be to check enrolments in the `help-to-save` mongo database before calling NS&I. That way it _would_ be an error that you'd want to be alerted about if a user who was enrolled our side didn't have an account on the NS&I side. Not something that has to be done in this story though, just an idea.

app/uk/gov/hmrc/helptosave/controllers/AccountController.scala
AndyHWChung: that's a valid point - this might be a question for Jean to see if we want to show something different to the user for our use-case when the information coming back from NS&I looks dodgy. The premise for returning a 500 here is that this endpoint in our BE is in charge of looking to see if the data coming back from NS&I looks OK and if so modify the format of the response to something more easy to use. If it turns out we do want to show different content if the data looks dodgy then we can go back and add a JSON body or something to the response so that our FE can pick that up.

app/uk/gov/hmrc/helptosave/connectors/HelpToSaveProxyConnector.scala
```diff
@@ -96,48 +110,64 @@ class HelpToSaveProxyConnectorImpl @Inject() (http:              WSHttp,
...
           response.status match {
             case Status.OK ⇒
-              val result = response.parseJson[NsiAccount].map(a ⇒ AccountO(Account(a)))
+              val result = response.parseJson[NsiAccount].flatMap(a ⇒
+                Account(a).toEither.bimap(s ⇒ s"[${s.toList.mkString(",")}]", Some(_)))
 
               result.fold(
                 e ⇒ {
-                  logger.warn(s"Could not parse getNsiAccount response, received 200 (OK), error=$e, response = ${maskNino(response.body)}", nino, correlationId)
+                  logger.warn(s"Could not parse getNsiAccount response, received 200 (OK), error=$e, response = ${maskNino(response.body)}",
```
frankoid: I don't think the response from NS&I should be logged to Kibana, even with the NINO masked, because it could contain other PII such as the individual's name & address.

test/uk/gov/hmrc/helptosave/connectors/HelpToSaveProxyConnectorSpec.scala
AndyHWChung: I might make sense to check our enrolment store before the call to NS&I - we can save an outbound call to NS&I if we see they don't have an account. The only case where this might be a problem is if the person has created an account but there has been a problem writing a record in our enrolment store for that user. Then if we're checking our enrolment store in this endpoint first, for this person we're blocking the call where it would have otherwise been successful. I think then, if we have the current behaviour where we don't check our enrolment store, then you're right - a pager duty alert is not needed in this case since it is kind of "expected" behaviour

frankoid:
I've had a quick look to check it matches what I think we agreed. Looks good in general, I think there is some logging of sensitive data that should be changed though (see my comments).

If you'd like me to do a more detailed review please let me know :)

app/uk/gov/hmrc/helptosave/connectors/HelpToSaveProxyConnector.scala
frankoid: I'm thinking of a situation where NS&I return a success response but we can't parse it (for example because some of the fields are missing or in the wrong format) - would this line be executed in that situation?

PR #140, opened 2018-06-11
[FD,BL] NGC-3251 Add Get Transactions endpoint

frankoid:
Also:
* Fix URL parameter escaping in get account.
* Don't log request body when JSON parsing fails in get account (in case the unparseable request body contains PII).
* Make logger.warn()s less noisy when JSON parsing fails in get account (previously HelpToSaveProxyConnector and AccountController would log similar things, now AccountController logs the error message returned by HelpToSaveProxyConnector).

### DIFF
--- a/app/uk/gov/hmrc/helptosave/connectors/HelpToSaveProxyConnector.scala
+++ b/app/uk/gov/hmrc/helptosave/connectors/HelpToSaveProxyConnector.scala
@@ -18,7 +18,7 @@ package uk.gov.hmrc.helptosave.connectors
 
 import java.util.UUID
 
-import cats.data.EitherT
+import cats.data.{EitherT, NonEmptyList}
 import cats.instances.int._
 import cats.instances.string._
 import cats.syntax.either._
@@ -124,18 +124,20 @@ class HelpToSaveProxyConnectorImpl @Inject() (http:              WSHttp,
           val _ = timerContext.stop()
           response.status match {
             case Status.OK ⇒
-              val result = response.parseJsonWithoutLoggingBody[NsiAccount].flatMap(a ⇒
-                Account(a).toEither.bimap(s ⇒ s"[${s.toList.mkString(",")}]", Some(_)))
+              val result = for {
+                nsiAccount ← response.parseJsonWithoutLoggingBody[NsiAccount].leftMap(NonEmptyList.one)
+                account ← Account(nsiAccount).toEither
+              } yield account
 
               result.bimap(
-                { e ⇒
+                { errors ⇒
                   metrics.getAccountErrorCounter.inc()
                   pagerDutyAlerting.alert("Could not parse JSON in the getAccount response")
-                  s"Could not parse getNsiAccount response, received 200 (OK), error=$e"
+                  s"Could not parse getNsiAccount response, received 200 (OK), error=[${errors.toList.mkString(",")}]"
                 },
                 { account ⇒
                   logger.info("Call to getNsiAccount successful", nino, "correlationId" → correlationId)
-                  account
+                  Some(account)
                 }
               )
 
@@ -177,19 +179,22 @@ class HelpToSaveProxyConnectorImpl @Inject() (http:              WSHttp,
           val _ = timerContext.stop()
           response.status match {
             case Status.OK ⇒
-              val result = response.parseJsonWithoutLoggingBody[NsiTransactions].flatMap(nsiTransactions ⇒
-                Transactions(nsiTransactions).toEither.bimap(s ⇒ s"[${s.toList.mkString(",")}]", Some(_)))
+              val result = for {
+                nsiTransactions ← response.parseJsonWithoutLoggingBody[NsiTransactions].leftMap(NonEmptyList.one)
+                transactions ← Transactions(nsiTransactions).toEither
+              } yield transactions
 
               result.bimap(
-                { e ⇒
+                { errors ⇒
                   metrics.getTransactionsErrorCounter.inc()
                   pagerDutyAlerting.alert("Could not parse get transactions response")
-                  s"Could not parse transactions response from NS&I, received 200 (OK), error=$e"
+                  s"""Could not parse transactions response from NS&I, received 200 (OK), error=[${errors.toList.mkString(",")}]"""
                 },
-                { transactions ⇒
+                { transactions: Transactions ⇒
                   logger.info("Call to get transactions successful", nino, "correlationId" → correlationId)
-                  transactions
-                })
+                  Some(transactions)
+                }
+              )
 
             case other ⇒
               if (isAccountDoesntExistResponse(response)) {
--- a/app/uk/gov/hmrc/helptosave/connectors/HelpToSaveProxyConnector.scala
+++ b/app/uk/gov/hmrc/helptosave/connectors/HelpToSaveProxyConnector.scala
@@ -183,7 +183,7 @@ class HelpToSaveProxyConnectorImpl @Inject() (http:              WSHttp,
               result.bimap(
                 { e ⇒
                   metrics.getTransactionsErrorCounter.inc()
-                  pagerDutyAlerting.alert("Could not parse JSON in the get transactions response")
+                  pagerDutyAlerting.alert("Could not parse get transactions response")
                   s"Could not parse transactions response from NS&I, received 200 (OK), error=$e"
                 },
                 { transactions ⇒
--- a/app/uk/gov/hmrc/helptosave/connectors/HelpToSaveProxyConnector.scala
+++ b/app/uk/gov/hmrc/helptosave/connectors/HelpToSaveProxyConnector.scala
@@ -23,19 +23,19 @@ import cats.instances.int._
 import cats.instances.string._
 import cats.syntax.either._
 import cats.syntax.eq._
-
 import com.google.inject.{ImplementedBy, Inject, Singleton}
+import io.lemonlabs.uri.dsl._
 import play.api.http.Status
 import play.api.libs.json.{Format, Json}
 import play.mvc.Http.Status.INTERNAL_SERVER_ERROR
 import uk.gov.hmrc.helptosave.config.{AppConfig, WSHttp}
 import uk.gov.hmrc.helptosave.metrics.Metrics
-import uk.gov.hmrc.helptosave.models.account.{Account, NsiAccount}
+import uk.gov.hmrc.helptosave.models.account.{Account, NsiAccount, NsiTransactions, Transactions}
 import uk.gov.hmrc.helptosave.models.{ErrorResponse, NSIUserInfo, UCResponse}
 import uk.gov.hmrc.helptosave.util.HeaderCarrierOps._
 import uk.gov.hmrc.helptosave.util.HttpResponseOps._
 import uk.gov.hmrc.helptosave.util.Logging._
-import uk.gov.hmrc.helptosave.util.{LogMessageTransformer, Logging, PagerDutyAlerting, Result, maskNino}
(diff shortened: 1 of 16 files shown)


build.sbt
```diff
@@ -101,11 +102,14 @@ lazy val wartRemoverSettings = {
   wartremoverErrors in(Compile, compile) ++= Warts.allBut(excludedWarts: _*)
 }
 
+lazy val catsSettings = scalacOptions += "-Ypartial-unification"
```
frankoid: I added this (and `scalaVersion := "2.11.12"`) in to avoid the need for type aliases to get the `.mapN` in the `Account` and `Transaction` validation to compile.

It did work, however IntelliJ still highlighed the `.mapN` calls in red without the type aliases so I left the type aliases in place. That means that this `-Ypartial-unification` is not strictly necessary for the code in this PR to compile, however I left in in place because it is recommended by https://typelevel.org/cats/

AndyHWChung: @frankoid thanks for that - creating type aliases here always gnawed at my side so it's good to be able to not use it. It's slightly annoying that intelliJ can't figure it out and shows an error though, but overall it's cleaner being able to not use the type aliases.

app/uk/gov/hmrc/helptosave/models/account/Transactions.scala
```diff
@@ -0,0 +1,151 @@
...
+  implicit val eqOperation: Eq[Operation] = Eq.fromUniversalEquals
+  implicit val writes: Writes[Operation] = new Writes[Operation] {
+    override def writes(o: Operation): JsValue = JsString(o match {
+      case Credit ⇒ "credit"
+      case Debit  ⇒ "debit"
+    })
+  }
+}
+
+case class ValidNsiTransaction(
```
AndyHWChung: This is only being used in the `apply` method in `Transactions`. To make it clearer that it's not for external use it'd be more obvious to make this case class unavailable outside of this file - perhaps as a private case class under in `Transactions`'s companion object

app/uk/gov/hmrc/helptosave/models/account/Transactions.scala
```diff
@@ -0,0 +1,151 @@
...
+  implicit val writes: Writes[Transactions] = Json.writes[Transactions]
+
+  private type ValidOrErrorString[A] = ValidatedNel[String, A]
+
+  def apply(nsiTransactions: NsiTransactions): ValidatedNel[String, Transactions] = {
+    nsiTransactions.transactions.toList
+      .traverse[ValidOrErrorString, ValidNsiTransaction](ValidNsiTransaction.apply)
+      .map(sortLikeNsiWeb)
+      .map(runningBalance)
+      .map(Transactions.apply)
```
AndyHWChung: you can avoid multiple `map` here by using function composition - it avoid having to look into the `Validated` more than once

app/uk/gov/hmrc/helptosave/models/account/Transactions.scala
```diff
@@ -0,0 +1,151 @@
...
+  }
+
+  private implicit val eqLocalDate: Eq[LocalDate] = Eq.fromUniversalEquals
```
AndyHWChung: you could provide a more meaningful `Eq` instance here using something like `Eq.instance[LocalDate](_.isEqual(_))`

app/uk/gov/hmrc/helptosave/models/account/Transactions.scala
```diff
@@ -0,0 +1,151 @@
...
+        else if (nsiOperation === "D") { Valid(Debit) }
+        else { Invalid(NonEmptyList.one(s"""Unknown value for operation: "$nsiOperation"""")) }
+
+      def sequenceValidation(nsiSequence: String): ValidOrErrorString[Int] =
+        try Valid(nsiSequence.toInt)
+        catch {
+          case _: NumberFormatException ⇒ Invalid(NonEmptyList.one(s"""Can't parse sequence value as an integer: "$nsiSequence""""))
+        }
+
+    (operationValidation(nsiTransaction.operation), sequenceValidation(nsiTransaction.sequence)).mapN {
```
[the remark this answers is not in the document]
frankoid: The reason we were trimming the account closed flag was because both `""` and `" "` are valid values. Those aren't valid values of `operation` so that reason for trimming doesn't apply here.

app/uk/gov/hmrc/helptosave/connectors/HelpToSaveProxyConnector.scala
```diff
@@ -159,9 +163,61 @@ class HelpToSaveProxyConnectorImpl @Inject() (http:              WSHttp,
...
+    EitherT[Future, String, Option[Transactions]](
+      http.get(url).map[Either[String, Option[Transactions]]] {
+        response ⇒
+          val _ = timerContext.stop()
+          response.status match {
+            case Status.OK ⇒
+              val result = response.parseJsonWithoutLoggingBody[NsiTransactions].flatMap(nsiTransactions ⇒
+                Transactions(nsiTransactions).toEither.bimap(s ⇒ s"[${s.toList.mkString(",")}]", Some(_)))
+
+              result.bimap(
```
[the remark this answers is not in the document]
frankoid: I tried to change this to bimap once but got an unwanted behaviour change (lack of transformation of the error message when `parseJsonWithoutLoggingBody` fails. See `NGC-3251-combine-bimaps` branch - `HelpToSaveProxyConnectorSpec` fails in that branch.
Perhaps we can have a look at the code together?

app/uk/gov/hmrc/helptosave/controllers/AccountQuery.scala
```diff
@@ -0,0 +1,54 @@
...
+import uk.gov.hmrc.helptosave.util
+import uk.gov.hmrc.helptosave.util.{LogMessageTransformer, Logging}
+import uk.gov.hmrc.helptosave.util.Logging._
+
+trait AccountQuery extends Logging with Results { this: HelpToSaveAuth ⇒
+
+  /**
+   * Behaviour common to actions that query for Help to Save account data based on NINO
+   */
+  protected def accountQuery[A](nino: String, systemId: String, correlationId: Option[String])(query: Request[AnyContent] ⇒ (String, String, String) ⇒ util.Result[Option[A]])(implicit transformer: LogMessageTransformer, writes: Writes[A]): Action[AnyContent] = authorised { implicit request ⇒ implicit authNino ⇒
```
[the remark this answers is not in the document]
frankoid: Done (I've added a case class)

app/uk/gov/hmrc/helptosave/models/account/Transactions.scala
frankoid: it needs to be accessible by `Transaction.apply(vnt: ValidNsiTransaction, balanceAfter: BigDecimal)` so perhaps we should just make `ValidNsiTransaction` and its companion object `private` instead of moving them inside `Transactions`?

app/uk/gov/hmrc/helptosave/models/account/Transactions.scala
frankoid: done - so I can learn from this what's the reason to do this? style? performance?

app/uk/gov/hmrc/helptosave/models/account/Transactions.scala
frankoid: done

app/uk/gov/hmrc/helptosave/connectors/HelpToSaveProxyConnector.scala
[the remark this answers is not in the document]
AndyHWChung: the problem might be that `parseJsonWithoutLoggingBody` returns `Either[String,NsiTransactions]` but `Transactions(nsiTransactions).toEither` returns   `Either[NonEmptyList[String],Transactions]`. The left types don't match up which means the flatmapping/bimapping might not work as you expect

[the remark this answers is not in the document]
frankoid: I managed to get the types to line up (i.e. write something that compiles), the problem is that the left-hand function in the bimap is not called if `parseJsonWithoutLoggingBody` returns a `Left` because `Either.flatMap` does not invoke the passed function on a `Left` only on a `Right`.

If there was such a thing as `Either.flatBimap` then we could use that but as it stands I can't think of a good way to avoid having 2 bimaps.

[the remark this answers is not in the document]
frankoid: I've got it working now by using `.left.map` - see https://github.com/hmrc/help-to-save/blob/NGC-3251-combine-bimaps/app/uk/gov/hmrc/helptosave/connectors/HelpToSaveProxyConnector.scala#L181

Let me know if you prefer that version, if so I'll apply that change to this branch and also do the same change in `getAccount`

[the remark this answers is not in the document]
AndyHWChung: I see what you mean now. The problem is the `bimap` is happening in the `flatMap`. The cleanest solution would probably be then to do the `flatMap` which builds up the result using the different functions used to build it up and then `bimap` on that result to handle the errors more consistently. Something like:
```scala
val result = for {
                nsiTransactions ← response.parseJsonWithoutLoggingBody[NsiTransactions].leftMap(NonEmptyList.one)
                transactions ← Transactions(nsiTransactions).toEither
              } yield {
                transactions
              }

              result.bimap(
                {
                  errors ⇒
                    metrics.getTransactionsErrorCounter.inc()
                    pagerDutyAlerting.alert("Could not parse get transactions response")
                    s"""Could not parse transactions response from NS&I, received 200 (OK), error=[${errors.toList.mkString(",")}]"""
                }, { transactions ⇒
                  logger.info("Call to get transactions successful", nino, "correlationId" → correlationId)
                  Some(transactions)
                }
              )
```
That's almost what you've done in  the `.left.map` implementation, but with the `for-comprehension` it's a bit easier to follow

app/uk/gov/hmrc/helptosave/models/account/Transactions.scala
AndyHWChung: efficiency - it becomes more apparent when you're doing multiple maps on something like a `Seq`, e.g.
```scala
 (1 to Int.MaxValue).map(_ + 1).map(_ * 2).map(_ - 3)
```
Here, each separate map requires a separate traversal of the `Seq` which is inefficient. With `Either` and `Validated` the hit isn't so dramatic as the example above, but in each map you're still having to check if it's a left or right in the background multiple times.

app/uk/gov/hmrc/helptosave/models/account/Transactions.scala
AndyHWChung: sounds good

app/uk/gov/hmrc/helptosave/connectors/HelpToSaveProxyConnector.scala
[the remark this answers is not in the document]
frankoid: Thanks, I've used your suggested implementation. There is a [slight change in the error string that ist returned in some circumstances](https://github.com/hmrc/help-to-save/pull/140/commits/ceeee8dae4dda12aefff3c5508ab2e9fc5c22347#diff-637ff59672f383b2286acaf4344f6b8bR495) but I think that's fine or even an improvement.

app/uk/gov/hmrc/helptosave/models/account/Transactions.scala
frankoid: done

AndyHWChung:
LGTM

PR #141, opened 2018-06-12
NGC-3251 Log unparseable response bodies when running locally 

frankoid:
Also set default port in build.sbt so there's no need to pass it to `sbt run`


AndyHWChung:
LGTM

PR #142, opened 2018-06-13
updated /create-account endpoint  and added /update-email to make it compatible with FE calls

sureshhmrc:


app/uk/gov/hmrc/helptosave/controllers/HelpToSaveController.scala
```diff
@@ -0,0 +1,101 @@
...
+class HelpToSaveController @Inject() (val enrolmentStore:          EnrolmentStore,
+                                      val itmpConnector:           ITMPEnrolmentConnector,
+                                      proxyConnector:              HelpToSaveProxyConnector,
+                                      userCapService:              UserCapService,
+                                      val eligibilityCheckService: EligibilityCheckService)(
+    implicit
+    transformer: LogMessageTransformer, appConfig: AppConfig)
+  extends BaseController with EligibilityBase with EnrolmentBehaviour with WithMdcExecutionContext {
+
+  def createAccount(): Action[AnyContent] = Action.async {
```
AndyHWChung: For these endpoints I think there is a minimum amount of authentication we can do which is to require the AuthProvider to be either `GovernmentGateway` or `PrivilegedApplication`.

AndyHWChung: e.g.
```scala
authorised(AuthProviders(GovernmentGateway, PrivilegedApplication)){ ... }
```

sureshhmrc: I thought about this, but how to check for the auth incase of stride requests ?

sureshhmrc: nvm, i just realised stride auth uses `PrivilegedApplication` type too , so is the same for API quests too ?

sureshhmrc: is it ok if i add auth to these endpoints in a upcoming PR  ?

AndyHWChung: yeah, stride and privileged access through the API would come under `PrivilegedApplication` - let's add the auth to your next PR then

AndyHWChung:
LGTM

PR #144, opened 2018-06-14
 HTS-1093: store eligibilityReason and channel in mongo and remove /enrol-user endpoint

sureshhmrc:


app/uk/gov/hmrc/helptosave/controllers/AccountQuery.scala
```diff
@@ -29,28 +29,32 @@ import uk.gov.hmrc.helptosave.util.{LogMessageTransformer, Logging}
...
-      val id = correlationId.getOrElse(UUID.randomUUID().toString)
-      query(request)(NsiAccountQueryParams(nino, systemId, id))
-        .fold(
-          { errorString ⇒
-            logger.warn(errorString, nino, "correlationId" → id)
-            InternalServerError
-          },
-          _.fold[Result](NotFound)(found ⇒ Ok(Json.toJson(found)))
-        )
+  protected def accountQuery[A](nino:          String,
```
sureshhmrc: nothing changed here in the file except the renaming `authorisedWithNino`

app/uk/gov/hmrc/helptosave/repo/EnrolmentStore.scala
```diff
@@ -144,11 +153,29 @@ class MongoEnrolmentStore @Inject() (mongo:   ReactiveMongoComponent,
...
+          case e ⇒
+            Left(e.getMessage)
+        }
+    )
 }
 
 object MongoEnrolmentStore {
 
-  private[repo] case class EnrolmentData(nino: String, itmpHtSFlag: Boolean)
+  private[repo] case class EnrolmentData(nino: String, itmpHtSFlag: Boolean, eligibilityReason: Option[Int] = None, channel: Option[String] = None)
```
sureshhmrc: this is used for reading from mongo and both fields are optional here just to make it compatible with existing records as those have only `nino` and `itmpFlag` fields

app/uk/gov/hmrc/helptosave/repo/EnrolmentStore.scala
```diff
@@ -96,12 +99,18 @@ class MongoEnrolmentStore @Inject() (mongo:   ReactiveMongoComponent,
...
+                             itmpFlag:          Boolean)(implicit ec: ExecutionContext): Future[WriteResult] =
+    collection.insert(BSONDocument("nino" -> nino, "itmpHtSFlag" -> itmpFlag, "eligibilityReason" -> eligibilityReason, "channel" -> channel))
+
   private[repo] def doUpdate(nino: NINO, itmpFlag: Boolean)(implicit ec: ExecutionContext): Future[Option[EnrolmentData]] =
     collection.findAndUpdate(
       BSONDocument("nino" -> nino),
       BSONDocument("$set" -> BSONDocument("itmpHtSFlag" -> itmpFlag)),
       fetchNewObject = true,
-      upsert         = true
+      upsert         = false
```
AndyHWChung: do we want `upsert=false`. The description for `upsert` is:
```
If set to true, creates a new document when no document matches the query
criteria. The default value is false, which does not insert a new document when no match is found.
```
In all cases when we're updating I think we'd expect a matching document to be present.

sureshhmrc: yes, thats why i didn't want to make it to `true` because incase if the document is not found for some reason then `true` makes it silently insert new record, but if we use `false` then it fails and we will get a chance to investigate the error 
And also we not making use of the returned record in the `true` case , then i don't see a point to to use `true` flag there

AndyHWChung: you're right - that makes sense

AndyHWChung:
LGTM

PR #149, opened 2018-06-19
 HTS-1093: Read stats from mongo for channel and eligibilityReason and log in kibana

sureshhmrc:


app/uk/gov/hmrc/helptosave/actors/EligibilityStatsActor.scala
```diff
@@ -0,0 +1,80 @@
...
+
+    case r: GetStatsResponse ⇒
+      eligibilityStatsHandler.handleStats(r.result)
+  }
+
+  def scheduleStats(): Cancellable = {
+    val scheduleStart = LocalTime.parse(config.getString("eligibility-stats.trigger-time"))
+    val timeUntilNextTrigger = timeCalculator.timeUntil(scheduleStart)
+    logger.info(s"Scheduling eligibility stats job in ${Time.nanosToPrettyString(timeUntilNextTrigger.toNanos)}")
+    scheduler.schedule(timeUntilNextTrigger, Duration(24, TimeUnit.HOURS), self, GetStats)
```
sureshhmrc: job runs everyday at 1.00 AM

app/uk/gov/hmrc/helptosave/repo/EnrolmentStore.scala
```diff
@@ -103,7 +103,7 @@ class MongoEnrolmentStore @Inject() (mongo:   ReactiveMongoComponent,
                              eligibilityReason: Option[Int],
                              source:            String,
                              itmpFlag:          Boolean)(implicit ec: ExecutionContext): Future[WriteResult] =
-    collection.insert(BSONDocument("nino" -> nino, "itmpHtSFlag" -> itmpFlag))
+    collection.insert(BSONDocument("nino" -> nino, "itmpHtSFlag" -> itmpFlag, "eligibilityReason" -> eligibilityReason, "source" -> source))
```
sureshhmrc: reenabling storing these fields in mongo

app/uk/gov/hmrc/helptosave/services/EligibilityStatsService.scala
```diff
@@ -0,0 +1,125 @@
...
+                                             metrics:                Metrics)(implicit ninoLogMessageTransformer: LogMessageTransformer)
+  extends EligibilityStatsService with Logging {
+
+  override def getEligibilityStats()(implicit ec: ExecutionContext): Future[Either[String, String]] = {
+    val timerContext = metrics.eligibilityStatsTimer.time()
+    eligibilityReportStore.getEligibilityStats()
+      .map { report ⇒
+        val _ = timerContext.stop()
+
+          def get(reason: Option[Int]): (Int, Int, Int, Int) = {
```
sureshhmrc: any better logic to get the individual metrics ?

app/uk/gov/hmrc/helptosave/repo/EligibilityStatsStore.scala
```diff
@@ -0,0 +1,75 @@
...
+  override def indexes: Seq[Index] = Seq(
+    Index(
+      key  = Seq("eligibilityReason" → IndexType.Ascending),
+      name = Some("eligibilityReasonIndex")
+    )
+  )
+
+  private[repo] def doAggregate(): Future[AggregationFramework.AggregationResult] = {
+    collection.aggregate(
+      Group(Json.obj("eligibilityReason" -> "$eligibilityReason", "source" -> "source"))("total" -> SumValue(1)),
```
AndyHWChung: Apparently `SumValue` is deprecated in favour of `SumAll`, or if `SumAll` isn't right then there's `Sum` or `SumField` as well

app/uk/gov/hmrc/helptosave/repo/EligibilityStatsStore.scala
```diff
@@ -0,0 +1,75 @@
...
+      List(Project(Json.obj("_id" -> 0, "eligibilityReason" -> "$_id.eligibilityReason", "source" -> "$_id.source", "total" -> "$total"))))
```
AndyHWChung: do we need the `"_id" -> 0` part? What is it doing here?

app/uk/gov/hmrc/helptosave/services/EligibilityStatsService.scala
```diff
@@ -0,0 +1,125 @@
...
+  def getEligibilityStats()(implicit ec: ExecutionContext): Future[Either[String, String]]
+
+}
+
+@Singleton
+class EligibilityStatsServiceImpl @Inject() (eligibilityReportStore: EligibilityStatsStore,
+                                             metrics:                Metrics)(implicit ninoLogMessageTransformer: LogMessageTransformer)
+  extends EligibilityStatsService with Logging {
+
+  override def getEligibilityStats()(implicit ec: ExecutionContext): Future[Either[String, String]] = {
```
AndyHWChung: In order to define the separation of concerns better, it makes more sense for this service to return [Future[EIther[String, List[EligibilityStats]]]` and for the `EligibilityStatsHandler` to look like:
```scala
trait EligibilityStatsHandler {
  def handleStats(result: List[EligibilityStats]): Unit
}
```
In this way, the service is in charge of getting the stats - or you may not even need the service if the actor can get an instance of the repo directly. And then the actor is in charge getting the stats and handling any errors and the handler is in charge of handling the stats (by logging them).

app/uk/gov/hmrc/helptosave/services/EligibilityStatsService.scala
AndyHWChung: rather than looking for specific values here, it'd be more general to make the service perhaps use `groupBy` on the list of `EligibilityStats` and group by eligibility reason, then channel. For example, something along the lines of:
```scala
case class X(a: Option[Int], b: Option[String], total: Int)
val l: List[X] =...
val groupedByA: Map[Option[Int], List[X]] =  l.groupBy(_.a)
val groupedByAThenB: List[(String, List[(String, Int)])] =
  groupedByA.map{ case (k,v) ⇒
    k.map(_.toString).getOrElse("Unknown") ->
      v.groupBy(_.b).map(x ⇒ x._1.getOrElse("Unknown") → x._2.map(_.total).sum).toList
  }.toList

```
Then you'd hav to have some logic to sort alphabetically and to insert the totals and to format the table correctly.

app/uk/gov/hmrc/helptosave/actors/EligibilityStatsActor.scala
AndyHWChung: it'd be nice to be able to run it perhaps at a more regular frequency - perhaps it'd be more flexible here to put in a list of times or an interval between jobs?

sureshhmrc: do we want to run this more than once a day ?  because it scans all mongo records in the collection, which can be a performance hit because if there are 3 instances in production there will be 3 such jobs at the moment, all running at the same time which is over doing for the same report actually...

app/uk/gov/hmrc/helptosave/repo/EligibilityStatsStore.scala
sureshhmrc: with out ```"_id" -> 0```  its returning different structure which makes more changes to parse the jsons out of the result , but if we have thi , we can straight away do 
```.map(_.head[EligibilityStats])```

File: app/uk/gov/hmrc/helptosave/actors/EligibilityStatsActor.scala (hunk 1 above)
Comment (AndyHWChung) [reply app/uk/gov/hmrc/helptosave/actors/EligibilityStatsActor.scala]: you've got a point there - we should probably be using the lock stuff to make sure that only one instance of this job is running per environment. We already have that lock stuff in the proxy so you should be able to re-use it. Having that in place, it'd be nice to perhaps log this information every 1 or 2 hours to be able to measure rates of enrolments

File: app/uk/gov/hmrc/helptosave/actors/EligibilityStatsActor.scala [hunk 7]
```diff
@@ -50,9 +50,10 @@ class EligibilityStatsActor(scheduler:               Scheduler,
 
   def scheduleStats(): Cancellable = {
     val scheduleStart = LocalTime.parse(config.getString("eligibility-stats.trigger-time"))
+    val frequency = config.get[FiniteDuration]("eligibility-stats.frequency").value
     val timeUntilNextTrigger = timeCalculator.timeUntil(scheduleStart)
     logger.info(s"Scheduling eligibility stats job in ${Time.nanosToPrettyString(timeUntilNextTrigger.toNanos)}")
-    scheduler.schedule(timeUntilNextTrigger, Duration(3, TimeUnit.HOURS), self, GetStats)
+    scheduler.schedule(timeUntilNextTrigger, frequency, self, GetStats)
```
sureshhmrc: added config for frequency of the job to every 3 hours,

app/uk/gov/hmrc/helptosave/actors/EligibilityStatsHandler.scala
```diff
@@ -0,0 +1,112 @@
...
+
+@ImplementedBy(classOf[EligibilityStatsHandlerImpl])
+trait EligibilityStatsHandler {
+  def handleStats(result: List[EligibilityStats]): String
+}
+
+@Singleton
+class EligibilityStatsHandlerImpl extends EligibilityStatsHandler with Logging {
+
+  override def handleStats(result: List[EligibilityStats]): String = {
```
sureshhmrc: this is the original logic i wrote, but i have come up with the following for dynamically formatting but it looks an overkill, so lets keep the  logic as it as ? 

```
@Singleton
class EligibilityStatsHandlerImpl extends EligibilityStatsHandler with Logging {

  val knownSources = Set("Digital", "Stride", "KCOM", "Unknown")

  override def handleStats(result: List[EligibilityStats]): String = {

    val groupedByEligibilityReason: Map[Option[Int], List[EligibilityStats]] = result.groupBy(_.eligibilityReason)
    val groupedByEligibilityReasonThenSource: List[(String, ArrayBuffer[(String, Int)])] =
      groupedByEligibilityReason.map {
        case (k, v) ⇒
          k.map(_.toString).getOrElse("Unknown") -> {
            val z = v.groupBy(_.source).map(x ⇒ x._1.getOrElse("Unknown") → x._2.map(_.total).sum).toList.sortBy(_._1)
            ArrayBuffer.concat(z)
          }
      }.toList.sortBy(_._1)

    //fill totals of missing sources with 0
    knownSources.foreach {
      source ⇒
        groupedByEligibilityReasonThenSource.foreach {
          case (p1, p2) ⇒
            var ml = mutable.ArrayBuffer.empty[(String, Int)]
            if (!p2.map(_._1).contains(source)) {
              val qq: (String, Int) = (source, 0)
              p2 += qq
            }
        }
    }

    //map integers to 6 ->  UC, 7 ->  WTC, 8 -> UC&WTC

    val newGroup = ArrayBuffer.empty[(String, ArrayBuffer[(String, Int)])]

    ArrayBuffer.concat(groupedByEligibilityReasonThenSource)
      .foreach {
        p ⇒
          if (p._1 === "6") {
            val t: (String, ArrayBuffer[(String, Int)]) = ("UC", ArrayBuffer.concat(p._2))
            newGroup += t
          } else if (p._1 === "7") {
            val t: (String, ArrayBuffer[(String, Int)]) = ("WTC", ArrayBuffer.concat(p._2))
            newGroup += t
          } else if (p._1 === "8") {
            val t: (String, ArrayBuffer[(String, Int)]) = ("UC&WTC", ArrayBuffer.concat(p._2))
            newGroup += t
          } else if (p._1 === "Unknown") {
            val t: (String, ArrayBuffer[(String, Int)]) = ("Unknown", ArrayBuffer.concat(p._2))
            newGroup += t
          }

      }

    logger.info(s"newgroup = $newGroup")

    val k = "|%8s|%10s|%10s|\n"

    val header: String =
      s"""
         |+--------+----------+----------+
         || Reason | Channel  |  Count   |
         |+--------+----------+----------+\n""".stripMargin

    var report = newGroup.foldLeft[String](header) {
      case (row: String, element: (String, ArrayBuffer[(String, Int)])) ⇒
        val a = element._2.foldLeft[String](row) {
          case (x: String, y: (String, Int)) ⇒
            x.concat(k.format(element._1, y._1, y._2))
        }

        def getTotalByReason =
          newGroup
            .filter(p ⇒ p._1 === element._1)
            .flatMap(_._2)
            .map(_._2)
            .sum

        a.concat(k.format("", "Total", getTotalByReason))
          .concat(k.format("", "", ""))

    }

    //compute totals by source
    knownSources.foreach {
      p =>
        val t = newGroup.map(_._2).map(_.toList.filter(_._1 === p).map(_._2)).map(_.sum).sum
        report = report.concat(k.format("Total", p, t))
    }

    //add grand total
    report = report
      .concat(k.format("", "Total", newGroup.map(_._2.map(_._2).sum).sum))
      .concat("+--------+----------+----------+")

    logger.info(s"report is $report")

    //for unit testing
    report
  }
```

AndyHWChung: I'd still recommend using a generic approach - when we add in new channels we don't want to keep remembering to update this report, and if there is a bug and there is some data written with a reason or channel we don't expect we would want to see it. Looking at the generic implementation above, it could perhaps be simplified by breaking our some of the logic into different methods. I notice there is also a lot of mutable state in there - we can avoid mutable state here by using folds and/or (tail) recursive functions to do the looping logic.

Perhaps to make it simpler, it would be easier to factor out the logic into something which looks like this:
```scala
// inner list is to support sub-keys and their values
type Table = List[(String,List[(String,String)])]

// handles gathering stats
def createTable(state: List[EligibilityStats]): Table

// handlers generic formatting logic
def prettyFormatTable(table: Table): String = ??? 
```
Then the logic would look something like:
```
val table: Table = // build table
val report: String = prettyFormatTable(table)
```
That way it's more easily understood what's going on

sureshhmrc: hm ok, just pushed the change, please check if anything can be improved

app/uk/gov/hmrc/helptosave/actors/EligibilityStatsHandler.scala
```diff
@@ -0,0 +1,148 @@
...
+            t :: modified
+          } else {
+            val t: (String, List[(String, Int)]) = original
+            t :: modified
+          }
+      }
+  }
+
+  def prettyFormatTable(table: Table): String = {
+    val f = "|%8s|%10s|%10s|\n"
```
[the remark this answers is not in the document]
sureshhmrc: wouldn't those strings be accommodated the specified length `%10s` ? 

if any string has more than 10 then we can adjust the table

app/uk/gov/hmrc/helptosave/actors/EligibilityStatsHandler.scala
```diff
@@ -107,42 +78,41 @@ class EligibilityStatsHandlerImpl extends EligibilityStatsHandler with Logging {
...
 
-          def getTotalByReason =
-            table
-              .filter(p ⇒ p._1 === element._1)
-              .flatMap(_._2)
-              .map(_._2)
-              .sum
+    var report = table.foldLeft[String](header) {
+      case (acc, curr) =>
+        val r = curr._2.foldLeft[String]("") {
```
sureshhmrc: its a list of list , not sure if we can create rows by using only one loop here (currently i used two `foldLeft`s ) ?

app/uk/gov/hmrc/helptosave/actors/EligibilityStatsActor.scala
```diff
@@ -0,0 +1,82 @@
...
+    case r: GetStatsResponse ⇒
+      eligibilityStatsHandler.handleStats(r.result)
+  }
+
+  def scheduleStats(): Cancellable = {
+    val scheduleStart = LocalTime.parse(config.getString("eligibility-stats.trigger-time"))
+    val frequency = config.get[FiniteDuration]("eligibility-stats.frequency").value
+    val timeUntilNextTrigger = timeCalculator.timeUntil(scheduleStart)
+    logger.info(s"Scheduling eligibility stats job in ${Time.nanosToPrettyString(timeUntilNextTrigger.toNanos)}")
+    scheduler.schedule(timeUntilNextTrigger, frequency, self, GetStats)
```
AndyHWChung: if the application starts just after the trigger time, we won't get any stats logged for a whole day - it would be better if the job was scheduled as soon as the actor started and then at the configurable set frequency thereafter to make sure we don't miss out on stats.

sureshhmrc: fixed!

AndyHWChung:
LGTM

PR #156, opened 2018-06-28
NGC-3264 Add openedYearMonth to get account

frankoid:
If you'd like to include bonusTerms.startDate as well that can easily be added by reverting the last commit in this PR.


AndyHWChung:
we currently have no need for startDate so we can just add that in if we need it in the future

AndyHWChung:
LGTM

PR #161, opened 2018-07-17
HTS-1160 - make sure gagues and registered more than once

AndyHWChung:


app/uk/gov/hmrc/helptosave/actors/EligibilityStatsActor.scala
```diff
@@ -61,12 +63,19 @@ class EligibilityStatsActor(scheduler:              Scheduler,
...
       case (reason, channels) ⇒
         channels.foreach {
           case (channel, _) ⇒
-            metrics.registerAccountStatsGauge(replaceSpaces(reason), replaceSpaces(channel),
-              () ⇒ statsTable.get(reason).flatMap(_.get(channel)).getOrElse(0))
+            if (!registeredStats.contains(reason → channel)) {
+              logger.info(s"Registering gauge for (reason, channel) = ($reason, $channel) ")
+              metrics.registerAccountStatsGauge(
+                replaceSpaces(reason), replaceSpaces(channel),
+                () ⇒ statsTable.get(reason).flatMap(_.get(channel)).getOrElse(0)
```
sureshhmrc: here you are still using `statsTable` is this expected ?

AndyHWChung: yes, the `statsTable` is supposed to be the single point of reference for the registered gauges to go to when working out what the counts should be

PR #162, opened 2018-07-25
[NGC-3394][BL] Remove accounting date sorting from transactions

howzat:
A bug in the initial implementation of the NS&I API meant that the sequence of
transactions reset to 1 at the beggining of each year. To retain the correct
ordering (as set out in the ICD) we had to additionally sort by the accounting
date. The bug is now marked as fixed so the sequence should be reliably
monotonic.


AndyHWChung:
LGTM

PR #163, opened 2018-07-27
HTS-1132 - added blocked from payments flag to account

AndyHWChung:
We would like to expose a flag which indicates whether or not payments into the account are blocked. The changes here are based on a query to NS&I on which blocking codes indicate such a state - their response was this table:

 Code | Description | Can user make payments?
  ---- | ---- | ---
  00 | No blocking | YES 
 11 | Block Applicant (i.e. Power) from withdrawing money | YES 
  12 | Block Applicant (i.e. Power) from paying money in | NO 
  13 | Block all Applicant (i.e. Power) permissions including Portal access | NO 
  15 | Block credits & Block Bonus Payments | NO 
  30 | Stop All | NO 
  61 | Stop all credits | NO 
  64 | Block all credits. Block all debits. Block Portal access | NO

### DIFF
--- a/app/uk/gov/hmrc/helptosave/models/account/Account.scala
+++ b/app/uk/gov/hmrc/helptosave/models/account/Account.scala
@@ -35,7 +35,7 @@ object BonusTerm {
   implicit val writes: Format[BonusTerm] = Json.format[BonusTerm]
 }
 
-case class Blocking(unspecified: Boolean)
+case class Blocking(unspecified: Boolean, payments: Boolean)
 
 object Blocking {
   implicit val writes: Format[Blocking] = Json.format[Blocking]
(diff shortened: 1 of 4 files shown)


app/uk/gov/hmrc/helptosave/models/account/Account.scala
```diff
@@ -108,9 +108,15 @@ object Account extends Logging {
...
-  private def nsiAccountToBlocking(nsiAccount: NsiAccount): Blocking = Blocking(
-    unspecified = nsiAccount.accountBlockingCode != "00" || nsiAccount.clientBlockingCode != "00"
-  )
+  private def nsiAccountToBlocking(nsiAccount: NsiAccount): Blocking = {
+      def isBlockedFromPredicate(predicate: String ⇒ Boolean): Boolean =
+        predicate(nsiAccount.accountBlockingCode) || predicate(nsiAccount.clientBlockingCode)
+
+    Blocking(
+      unspecified = isBlockedFromPredicate(_ =!= "00"),
+      payments    = isBlockedFromPredicate(s ⇒ s =!= "00" && s =!= "11")
```
frankoid: This will return Blocking.payments = true for unknown blocking codes. That seems reasonable (failsafe) but perhaps it would be worth logging unknown blocking codes when they are encountered.

AndyHWChung: added logging now for unknown blocking codes - I've implemented behaviour where an error is returned if there are unknown blocking codes in alignment with the behaviour with other fields when unknown values are passed onto us

PR #166, opened 2018-08-06
[NGC-3584][FD] Add accountHolderName and accountHolderEmail account fields

frankoid:


frankoid:
Superseded by #167 which doesn't combine the name fields (that task will be left to mobile-specific code)

PR #168, opened 2018-08-21
HTS-1229 - added logging for stride manual accounts

AndyHWChung:


AndyHWChung:
see also https://github.com/hmrc/help-to-save-stride-frontend/pull/97

PR #171, opened 2018-08-23
HTS-1248 corrected query parameters

AndyHWChung:


test/uk/gov/hmrc/helptosave/connectors/HelpToSaveProxyConnectorSpec.scala
```diff
@@ -518,45 +508,12 @@ class HelpToSaveProxyConnectorSpec extends TestSupport with MockPagerDuty with E
...
           """.stripMargin)
 
-        mockGet(getTransactionsUrl)(Some(HttpResponse(400, Some(errorResponse))))
+        mockGet(getTransactionsUrl, queryParameters)(Some(HttpResponse(400, Some(errorResponse))))
 
         val result = await(proxyConnector.getTransactions(nino, systemId, correlationId).value)
         result shouldBe Right(None)
       }
 
-      "handle characters in correlationId and systemId that need to be escaped" in {
```
AndyHWChung: @frankoid  I deleted this test and a similar one above because it doesn't look like we need it anymore. I've just refactored the code to use the injectable `HttpClient` from the bootstrap library. It has a `GET` method where you supply the query parameters of the request and does the URL encoding for you (see `makeQueryString` here: https://github.com/hmrc/http-verbs/blob/master/src/main/scala/uk/gov/hmrc/http/HttpGet.scala#L36). Are you happy for these tests to be deleted?

frankoid: Good spot, I'll be using that feature in future!

Yes, we can trust that the library does its job and escapes query parameters, all we have to do here is check we're calling it correctly - which `mockGet(getAccountUrl, queryParameters)` does (it ensures that the query parameters are being passed as a separate Map not as part of the query string).

PR #173, opened 2018-08-24
HTS-1248 - corrected last remaining query parameters

AndyHWChung:


app/uk/gov/hmrc/helptosave/connectors/DESConnector.scala
```diff
@@ -57,20 +57,21 @@ class DESConnectorImpl @Inject() (http: HttpClient)(implicit transformer: LogMes
 
   val originatorIdHeader: (String, String) = "Originator-Id" → appConfig.getString("microservice.services.paye-personal-details.originatorId")
 
-  private def url(nino: String, ucResponse: Option[UCResponse]): String = {
+  def eligibilityCheckUrl(nino: String): String = s"$itmpECBaseURL/help-to-save/eligibility-check/$nino"
```
sureshhmrc: in the AccountControllerSpec, i see
`val fakeRequest = FakeRequest("GET", s"/nsi-account?$queryString")` does it needs to be changes as well ?

sureshhmrc: and in `TransactionsControllerSpec` too

AndyHWChung: no, we don't even need the URL in the fake request really - it'll work without it

AndyHWChung: that stuff doesn't go through the injectable `HttpClient`

PR #176, opened 2018-09-04
HTS-1279 Added manually entered BankDetails to AccountCreated event

Jackie1988:


app/uk/gov/hmrc/helptosave/models/HTSEvent.scala
```diff
@@ -18,85 +18,144 @@ package uk.gov.hmrc.helptosave.models
...
+      userInfo.contactDetails.phoneNumber.getOrElse(""),
+      userInfo.contactDetails.communicationPreference,
+      userInfo.registrationChannel,
+      source
     ),
+                           Some(ManuallyEnteredDetails(
+        userInfo.nbaDetails.fold("")(_.accountName),
+        userInfo.nbaDetails.fold("")(_.accountNumber),
+        userInfo.nbaDetails.fold("")(_.sortCode),
+        userInfo.nbaDetails.fold("")(_.rollNumber.getOrElse(""))
```
AndyHWChung: to avoid both the `fold` and `getOrElse` you could do `userInfo.nbaDetails.flatMap(_.rollNumber).getOrElse("")

app/uk/gov/hmrc/helptosave/models/HTSEvent.scala
```diff
@@ -18,85 +18,144 @@ package uk.gov.hmrc.helptosave.models
...
+      userInfo.contactDetails.postcode,
+      userInfo.contactDetails.countryCode.getOrElse(""),
+      userInfo.contactDetails.email.getOrElse(""),
+      userInfo.contactDetails.phoneNumber.getOrElse(""),
+      userInfo.contactDetails.communicationPreference,
+      userInfo.registrationChannel,
+      source
     ),
+                           Some(ManuallyEnteredDetails(
+        userInfo.nbaDetails.fold("")(_.accountName),
```
AndyHWChung: for consistency with the above code `getOrElse` would look neater - or changing the above `getOrElse`s to `fold`s

app/uk/gov/hmrc/helptosave/models/HTSEvent.scala
```diff
@@ -18,85 +18,144 @@ package uk.gov.hmrc.helptosave.models
...
             s"resultCode=${eligibilityResult.resultCode}, reasonCode=${eligibilityResult.reasonCode}, " +
             s"meaning result='${eligibilityResult.result}', reason='${eligibilityResult.reason}'"
-
-          Map[String, String]("nino" → nino, "eligible" → "false", "reason" -> reason)
-
+          EligibilityResult(nino, false, Some(reason))
         }
-      result ++ ucData(ucResponse)
+
+      ucData(result, ucResponse)
```
AndyHWChung: what might be neater here is to have an apply method on `EligibilityResult` which looks like this:
```scala
object EligibilityResult{
  def apply(nino: String,  eligibilityResult: EligibilityCheckResult, ucResponse: Option[UCResponse])
}
```
rather than half constructing the case class with default values

app/uk/gov/hmrc/helptosave/models/HTSEvent.scala
Jackie1988: ok thanks

app/uk/gov/hmrc/helptosave/models/HTSEvent.scala
Jackie1988: as nbaDetails is an Option, how do I access the nested values with 'getOrElse' ?

AndyHWChung: oh, that's right, I forgot it's an option. In that case you could leave it as you have coded it now. Alternatively there's `userInfo.nbaDetails.map(_.accountName).getOrElse("")` but that's a but more long-winded

app/uk/gov/hmrc/helptosave/models/HTSEvent.scala
Jackie1988: ok I've made changes to that section, thanks

app/uk/gov/hmrc/helptosave/models/HTSEvent.scala
Jackie1988: ok done, thanks

AndyHWChung:
LGTM

PR #192, opened 2018-10-25
HTS-1208: Log Get Account response from NS&I in Splunk

sureshhmrc:


app/uk/gov/hmrc/helptosave/connectors/HelpToSaveProxyConnector.scala
```diff
@@ -142,6 +144,10 @@ class HelpToSaveProxyConnectorImpl @Inject() (http:              HttpClient,
             val _ = timerContext.stop()
             response.status match {
               case Status.OK ⇒
+
+                val path = s"/help-to-save/$nino/account?nino=$nino&systemId=$systemId&correlationId=$correlationId"
```
sureshhmrc: note that this is the url of the incoming request but not the NSI get account url

AndyHWChung: it'll be more futureproof to pass the path in and get it from the `request` in the controller - you can find the full URI including the query parameters under `request.uri`

sureshhmrc: fixed

app/uk/gov/hmrc/helptosave/models/HTSEvent.scala
```diff
@@ -154,3 +154,16 @@ object AccountCreated {
...
+
+case class GetAccountResult(nino: String, account: JsValue)
+
+object GetAccountResult {
+  implicit val format: Format[GetAccountResult] = Json.format[GetAccountResult]
+}
+
+case class GetAccountResultEvent(getAccountResult: GetAccountResult, path: String)(implicit hc: HeaderCarrier, appConfig: AppConfig) extends HTSEvent {
+  val value: ExtendedDataEvent = {
+    HTSEvent(appConfig.appName, "GetAccountResult", Json.toJson(getAccountResult), "GetAccountResult", path)
```
AndyHWChung: to keep the transaction name consistent with the others, could we have "get-account-result" for the transaction name here?

app/uk/gov/hmrc/helptosave/models/HTSEvent.scala
```diff
@@ -154,3 +154,16 @@ object AccountCreated {
...
 }
+
+case class GetAccountResult(nino: String, account: JsValue)
+
+object GetAccountResult {
+  implicit val format: Format[GetAccountResult] = Json.format[GetAccountResult]
+}
+
+case class GetAccountResultEvent(getAccountResult: GetAccountResult, path: String)(implicit hc: HeaderCarrier, appConfig: AppConfig) extends HTSEvent {
+  val value: ExtendedDataEvent = {
```
AndyHWChung: there's a `HtsEventSpec` you can add to to add a unit test to check that the fields are written as expected

app/uk/gov/hmrc/helptosave/models/HTSEvent.scala
sureshhmrc: fixed

app/uk/gov/hmrc/helptosave/models/HTSEvent.scala
sureshhmrc: added unit test

AndyHWChung:
LGTM

PR #194, opened 2018-10-26
HTS-1137 Added threshold to the eligibility check response to send to FE

Jackie1988:


app/uk/gov/hmrc/helptosave/models/EligibilityCheckResult.scala
```diff
@@ -28,3 +28,11 @@ object EligibilityCheckResult {
   implicit val format: Format[EligibilityCheckResult] = Json.format[EligibilityCheckResult]
 
 }
+
+case class EligibilityCheckResponse(eligibilityCheckResult: EligibilityCheckResult, threshold: Option[Double])
```
sureshhmrc: does the `threshold` has to be option ? are there any cases where we miss this value ?

PR #195, opened 2018-11-07
Prevented PAYEPersonDetails JSON validation throwing errors

AndyHWChung:


app/uk/gov/hmrc/helptosave/models/PayePersonalDetails.scala
```diff
@@ -102,30 +102,29 @@ object PayePersonalDetails {
...
-          JsSuccess(Address(line1, line2, line3, line4, line5, postcode, countryCode.flatMap(countryCodes.get).map(_.take(2))))
-        }
+      )(v ⇒
+          for {
+            line1 ← (v \ "line1").validate[String]
+            line2 ← (v \ "line2").validate[String]
+            line3 ← (v \ "line3").validateOpt[String]
+            line4 ← (v \ "line4").validateOpt[String]
+            line5 ← (v \ "line5").validateOpt[String]
+            postcode ← (v \ "postcode").validate[String]
```
sureshhmrc: would this postcode be evaluated if the options line3, line4, or line5 are not present ?

AndyHWChung: yes, the `flatMap` is being done on  `JsResult` rather than on `Option`

sureshhmrc: ah ok , no issues then

PR #196, opened 2018-11-12
HTS-1334: Moved Bars logic from frontend

sureshhmrc:


app/uk/gov/hmrc/helptosave/controllers/HelpToSaveController.scala
```diff
@@ -105,6 +106,26 @@ class HelpToSaveController @Inject() (val enrolmentStore:         EnrolmentStore
     implicit request ⇒ checkEligibility(nino, routes.HelpToSaveController.checkEligibility(nino).url)
   }
 
+  def doBarsCheck(): Action[AnyContent] = ggOrPrivilegedAuthorised { implicit request ⇒
+    request.body.asJson.map(_.validate[BarsRequest]) match {
+
+      case Some(JsSuccess(barsRequest, _)) ⇒
+        barsService.validate(barsRequest, request.uri).flatMap {
```
sureshhmrc: note that the `request.uri` is always the BE uri `/help-to-save/validate-bank-details` irrespective of whether request originates from FE or API layer

app/uk/gov/hmrc/helptosave/models/BankDetails.scala
```diff
@@ -18,8 +18,14 @@ package uk.gov.hmrc.helptosave.models
...
 import play.api.libs.json.{Format, Json}
 
+case class BankDetails(sortCode: String, accountNumber: String, rollNumber: Option[String], accountName: String)
+
 object BankDetails {
   implicit val bankDetailsFormat: Format[BankDetails] = Json.format[BankDetails]
 }
 
-case class BankDetails(sortCode: String, accountNumber: String, rollNumber: Option[String], accountName: String)
+case class BarsRequest(nino: String, sortCode: String, accountNumber: String)
```
sureshhmrc: reading `nino` from the caller instead of retrieving it from Auth - this is due to API requests auth session may not always contain the `nino`

app/uk/gov/hmrc/helptosave/controllers/HelpToSaveController.scala
sureshhmrc: i mean its the path param for auditing

app/uk/gov/hmrc/helptosave/services/BarsService.scala
```diff
@@ -0,0 +1,98 @@
...
+import uk.gov.hmrc.http.HeaderCarrier
+
+import scala.concurrent.{ExecutionContext, Future}
+
+@ImplementedBy(classOf[BarsServiceImpl])
+trait BarsService {
+
+  type BarsResponseType = Future[Either[String, Boolean]]
+
+  def validate(barsRequest: BarsRequest, path: String)(implicit hc: HeaderCarrier, ec: ExecutionContext): BarsResponseType
```
AndyHWChung: you should be able t get the path from the header carrier somewhere - you shouldn't need to pass it in as a separate parameter here

sureshhmrc: we can get it from `Request` but not sure if its a good idea to deal with Request entities in the services layer , but i can do it

AndyHWChung: if that's the case, there's no point in making that change as we'll just be swapping one input for another. It's just slightly annoying the info isn't there in the HeaderCarrier which we're passing in anyway

AndyHWChung:
LGTM

PR #201, opened 2018-11-29
HTS-1490: Added secure stride role to StrideAuth

Jackie1988:


conf/application.conf
```diff
@@ -215,41 +215,41 @@ crypto.encryption-key = "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA="
...
-  }
-  get-transactions {
-    version = "V1.0"
-  }
-  create-account {
-    version = "V2.0"
-  }
+stride {
+  base64-encoded-roles = [ "aHRzIGhlbHBkZXNrIGFkdmlzb3I=" ]
+  base64-encoded-secure-roles = [ "aHRzIGhlbHBkZXNrIGFkdmlzb3Igc2VjdXJl" ]
```
sureshhmrc: why do have to maintain two different roles configs ? 
can't we just maintain in a single config  with `comma` separated list of roles in them 
`base64-encoded-roles = ["hts helpdesk advisor", "hts helpdesk advisor secure"]`

AndyHWChung: we need to be able to check if the call operator has either one set of roles or the other. So we want to be able to do the following:
```scala
val standardRoles = List("a", "b")
val secureRoles = List("c", "d")

List("a"), List("a", "c"), List("d"), etc. // don't allow access
List("a", "b"), List("c", "d") // allow access
```
We wouldn't be able to exclude the `List("a", "c") case with a single list in the config

PR #203, opened 2018-12-18
HTS-1546 - upgraded to bootstrap 4.4.0

AndyHWChung:


app/uk/gov/hmrc/helptosave/controllers/HelpToSaveAuth.scala
```diff
@@ -48,7 +48,7 @@ class HelpToSaveAuth(htsAuthConnector: AuthConnector) extends BaseController wit
   private type HtsAction = Request[AnyContent] ⇒ Future[Result]
   private type HtsActionWithNino = Request[AnyContent] ⇒ NINO ⇒ Future[Result]
 
-  def ggAuthorisedWithNino(action: HtsActionWithNino): Action[AnyContent] =
+  def ggAuthorisedWithNino(action: HtsActionWithNino)(implicit ec: ExecutionContext): Action[AnyContent] =
```
sureshhmrc: can we move this `implicit ec` declaration to to the top of the class, instead of keeping it at method level ?

PR #205, opened 2019-02-01
Hts 1592 Find enrolment and Get email mongo queries ignore nino suffix

Jackie1988:


app/uk/gov/hmrc/helptosave/repo/EmailStore.scala
```diff
@@ -132,13 +136,12 @@ class MongoEmailStore @Inject() (mongo:   ReactiveMongoComponent,
...
   }
 
-  private[repo] def doUpdate(encryptedEmail: String, nino: NINO)(implicit ec: ExecutionContext): Future[Option[EmailData]] =
-    collection.findAndUpdate(
-      BSONDocument("nino" -> nino),
-      BSONDocument("$set" -> BSONDocument("email" -> encryptedEmail)),
-      fetchNewObject = true,
-      upsert         = true
-    ).map(_.result[EmailData])
+  private[repo] def doUpdate(encryptedEmail: String, nino: NINO)(implicit ec: ExecutionContext): Future[Boolean] =
```
AndyHWChung: would we need to delete by regex when deleting emails as well?

app/uk/gov/hmrc/helptosave/repo/EnrolmentStore.scala
```diff
@@ -117,7 +117,9 @@ class MongoEnrolmentStore @Inject() (mongo:   ReactiveMongoComponent,
     {
       val timerContext = metrics.enrolmentStoreGetTimer.time()
 
-      find("nino" → JsString(nino)).map { res ⇒
+      val regex = "^" + nino.take(8) + ".$"
```
AndyHWChung: would we need to update by regex when updating the enrolment store?

app/uk/gov/hmrc/helptosave/repo/EmailStore.scala
Jackie1988: Yes as its possible it won't be picked up if a different suffix is used

AndyHWChung:
LGTM

PR #226, opened 2019-10-22
Dls 813

JackieDev:


conf/application.conf
```diff
@@ -57,7 +57,7 @@ play.modules.enabled += "uk.gov.hmrc.helptosave.modules.EligibilityStatsModule"
 # Global request handler
 # ~~~~
 # Set to the object handler for migrated 2.3 services
-play.http.requestHandler = "play.api.http.GlobalSettingsHttpRequestHandler"
+#play.http.requestHandler = "play.api.http.GlobalSettingsHttpRequestHandler"
```
lutnos: Better to remove that section than to comment a line, I think

Jackie1988: Think that may be a mistake, I forgot to put it back after commenting it out. Will put it back in and see what happens.

Jackie1988: It looks like the GlobalSettingsHttpRequestHandler has disappeared so I have updated it to JavaCompatibleHttpRequestHandler which extends DefaultHttpRequestHandler. https://github.com/playframework/playframework/issues/5357

Jackie1988: I realize the above link is a bit old (Dec 2015)

PR #236, opened 2020-08-17
Dls 3062 - help-to-save can leek PII when bad address data FIX

gavinharris43:


app/uk/gov/hmrc/helptosave/models/PayePersonalDetails.scala
```diff
@@ -104,15 +105,15 @@ object PayePersonalDetails {
...
       )(v ⇒
           for {
-            line1 ← (v \ "line1").validate[String]
-            line2 ← (v \ "line2").validate[String]
-            line3 ← (v \ "line3").validateOpt[String]
-            line4 ← (v \ "line4").validateOpt[String]
-            line5 ← (v \ "line5").validateOpt[String]
-            postcode ← (v \ "postcode").validate[String]
-            countryCode ← (v \ "countryCode").validateOpt[Int]
+            line1 ← lookup("line1", v).validate[String]
```
lutnos, line 108: You ought to call this method in the lookups for Name and DOB, otherwise an error in either of those would cause the same problem, I think

app/uk/gov/hmrc/helptosave/util/JsLookupHelper.scala
```diff
@@ -0,0 +1,34 @@
...
+
+  /**
+   * Return the property corresponding to the fieldName, supposing we have a JsObject.
+   *
+   * @param fieldName the name of the property to look up
+   */
+  def lookup(fieldName: String, value: JsValue): JsLookupResult  = JsDefined(value) match {
+    case JsDefined(obj @ JsObject(_)) =>
+      obj.value.get(fieldName).map(JsDefined.apply)
+        .getOrElse(JsUndefined(s"'$fieldName' is undefined on object: ${obj.keys}"))
```
lutnos, line 15: It would be nicer to do say: ${obj.keys.map(s"'${_}'").mkString(","}, to be consistent with the rest of the error message string

app/uk/gov/hmrc/helptosave/models/PayePersonalDetails.scala
gavinharris43: Had a look at Name and DOB, don't leak PII if there is any missing data, as implemented in a different way, so current implementation can be as is.

app/uk/gov/hmrc/helptosave/util/JsLookupHelper.scala
gavinharris43: done

lutnos (approved):

PR #245, opened 2021-12-02
[DLS-4700][JR]returning account details to Account call

JakeReid2020:


lutnos (commented):
Why do you want as PR for such a small part of the change?  Is it because the NGC team need it?

PR #252, opened 2023-01-09
[SM][DLS-6603] Replaced hmrc-lock with hmrc-mongo-lock, also removed …

SeanMurray1:
…mongo-lock dependency. Due to this service using a release lock we couldn't use the TimePeriodLockService because it didn't contain a function to release the lock.

Signed-off-by: Sean Murray <sean.murray@digital.hmrc.gov.uk>


rambabu-posa (approved):
LGTM

PR #253, opened 2023-01-26
Dls 6514 fix

Chrissystk:


lutnos (changes requested):
Cheers Chrissy, but please could we have the copyright change dates in a separate commit.  Otherwise, both for the PR and for the eventual git history it can be very difficult to find the functional change needle in the copyright date change haystack. Thanks

lutnos (approved):
Approved, now that https://github.com/hmrc/help-to-save/pull/254 has been merged

PR #264, opened 2023-06-13
Use syntax instead of functions

LAJW:
An experimental refactor PR showing how we could structure the code with basic language constructs to make it easier to read and debug.

Reasons to use match instead of .fold`:
* Easy to confuse whether we're handling positive or negative case
* Difficult to compose with pattern matching
* Difficult to chain with `for` expressions which leads to nesting
* Easier to debug (easier to select the handling case, keeps the stack flat)

Reasons to use `for` instead of `.flatMap`/`.map`:
* Easier to debug (procedural-looking code)
* Encourages chaining (with EitherT)
* Less nesting

Reasons to use `.pipe` / `.tap`:
* Avoids creating unnecessary variables (restricts scope of values)
* Allows you to chain outside of monads


lutnos (commented):
I strongly support this kind of improvement, yes

TomRafferty:
closed due to inactivity

PR #267, opened 2023-12-09
DLS-9402 | Fix partial accounts by introducing soft deletion of enrolments

mahesh-hmrc:
PR to address partial accounts that are blocking users to enrol again, after their accounts have been deleted from downstream systems. Reasons for this has been explained in detail with comments and linked ticket under https://jira.tools.tax.service.gov.uk/browse/DLS-9402

Changes in this PR are a step in direction where we address deletions in alignment with when the deletions happen in downstream systems. Till then, we need a mechanism to unblock users and soft delete mechanism in this PR paves way towards the final change. Reasons for using deleteFlag has been mentioned in README and also here.

Changes might look like they need lot of testing, but the workflow is simple enough. Any other approach would require the same amount of testing, with the added risk that we could've mutated the key field like `nino`. Which is generally not recommended and deleteFlag pattern being followed for soft deletes.

New test scenarios added to the store cover the different cases that we have taken into consideration for both deletions and undoing deletions.

Config processing is split into 2 stages for deletions and undoing deletions because of following reason:

Though I personally don't see a valid reason we would undo an enrolment when the source of that in downstream has been deleted. Logic has been included based on ask/discussions we've had around this. Because, data lineage with downstream systems is always abt looking ahead rather than mutating previous state. 

However, consider that undoing is a valid requirement and business usecase. Then, there will be a case of undoing deletions where the user has an active enrolment with 1 or more previous soft delete tagged enrolments. And, we now want to undo deletion for one of the 2 enrolments that are soft deleted. In that case, we can't risk undoing deletion without marking the current one as deleted, otherwise it would present us with multiple enrolments that are active. 

So, to make the process simple it would be ideal to have separate config processing for deletions and undoing them, as they can be sequentially done by following config

```
enrolment.delete-ninos=[{"nino": "nino-to-delete"}]
enrolment.undo-delete-ninos=[{"nino": "nino-to-delete", "docId": "doc-id-of-enrolement-version-to-become-active"}]

```

As these actions are processed sequentially, at end of the run, there will only be a single active enrolment against that nino. Approach to rely on doc id puts a dependency on knowing which of the enrolments we want to undo deletion for and make it active.This step of relying on access to live data is anyway needed with other alternate approaches (like prefixing NINOs). Because, prefixes doesn't have context on the enrolment and to know which prefix (with an infix increment) needs to be reverted involves assessing live data and then including that in config.

### DIFF
--- a/app/uk/gov/hmrc/helptosave/services/ApplicationStart.scala
+++ b/app/uk/gov/hmrc/helptosave/services/ApplicationStart.scala
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2023 HM Revenue & Customs
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.helptosave.services
+
+import play.api.Logging
+import uk.gov.hmrc.helptosave.config.AppConfig
+import uk.gov.hmrc.helptosave.models.NINODeletionConfig
+import uk.gov.hmrc.helptosave.repo.EnrolmentStore
+
+import javax.inject.{Inject, Singleton}
+import scala.concurrent.ExecutionContext
+import scala.util.{Failure, Success}
+
+@Singleton
+class ApplicationStart @Inject() (appConfig: AppConfig, enrolmentStore: EnrolmentStore, executionContext: ExecutionContext) extends Logging {
+
+  implicit val ec: ExecutionContext = executionContext
+
+  val ninosToDelete: Seq[NINODeletionConfig] = appConfig.ninoDeletionConfig("delete-ninos")
+  if (ninosToDelete.nonEmpty) {
+    enrolmentStore.updateDeleteFlag(ninosToDelete).value.onComplete {
+      case Success(Right(_))       => logger.info(s"Successfully deleted list of NINOs : $ninosToDelete")
+      case Success(Left(errorMsg)) => logger.error(errorMsg)
+      case Failure(ex)             => logger.error(s"Failed to delete configured list of NINOs: $ninosToDelete", ex)
+    }
+  }
+
+  val ninosToUndoDeletion: Seq[NINODeletionConfig] = appConfig.ninoDeletionConfig("undo-delete-ninos")
+  if (ninosToUndoDeletion.nonEmpty) {
+    enrolmentStore.updateDeleteFlag(ninosToUndoDeletion, revertSoftDelete = true).value.onComplete {
+      case Success(Right(_))       => logger.info(s"Successfully undid deletion of NINOs, $ninosToUndoDeletion")
+      case Success(Left(errorMsg)) => logger.error(errorMsg)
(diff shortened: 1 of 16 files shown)


lutnos (commented):
We need to generate audit events for each change.  The undo delete operation needs document IDs and we currently have no way of obtaining them.  The audit event ought to contain the IDs of all documents updated by an undelete, this could then be used when undoing deletes.

There are also untidy ends around the check on NINOs, for example when undoing a delete, the check only makes sure that something with that NINO exists, not necessarily the right documentID.

Again the document IDs from the BulkWriteResult are not sufficient, because we dio not know which are associated with which NINO.

I would suggest using the initial find to produce a complete list of NINOs with document IDs.  That could perhaps be audited.  Then best to bulk update using that list.  Do we then need to check the BulkWriteResult for any possible partial success - I am not clear.

I started an implementation of this myself, FWIW, but it got bogged down in converting an EitherT.  This may not be the best approach anyway.  On branch dave/DLS-9402/audit-soft-deletes, does not compile

lutnos (commented):
@mahesh-hmrc 

It should be quite simple to audit undo of delete, because the document ID is present in the config.

But for delete, it is more tricky.  What I would suggest is to use the initial find() to augment the config with document Ids.  The new "config" with IDs can be passed out for delete instead.  At that point it is no longer the same Seq[NINODeletionConfig] type, rather it should be a type of something like Seq((NINO,ObjectId)).

That is unless you know an easier way, the bulk write can return document IDs but I am not sure that can be tied back to NINOs

mahesh-hmrc:
Thanks Dave! Yeah, think that's inline with what we discussed earlier and is already being factored into the models

app/uk/gov/hmrc/helptosave/services/ApplicationStart.scala
```diff
@@ -0,0 +1,70 @@
...
+                                  executionContext: ExecutionContext,
+                                  audit:            HTSAuditor) extends Logging {
+
+  implicit val ec: ExecutionContext = executionContext
+
+  val ninosToDelete: Seq[NINODeletionConfig] = appConfig.ninoDeletionConfig("delete-ninos")
+  if (ninosToDelete.nonEmpty) {
+    enrolmentStore.updateDeleteFlag(ninosToDelete).value.onComplete {
+      case Success(Right(deletedAccounts)) =>
+        val ninos = deletedAccounts.map(_.nino).toList
```
lutnos, line 43: We need to log not just the NINOs here, but the _id from EnrolmentData .

app/uk/gov/hmrc/helptosave/services/ApplicationStart.scala
```diff
@@ -0,0 +1,70 @@
...
+      case Success(Left(errorMsg)) => logger.error(errorMsg)
+      case Failure(ex)             => logger.error(s"Failed to delete configured list of NINOs: $ninosToDelete", ex)
+    }
+  }
+
+  val ninosToUndoDeletion: Seq[NINODeletionConfig] = appConfig.ninoDeletionConfig("undo-delete-ninos")
+  if (ninosToUndoDeletion.nonEmpty) {
+    enrolmentStore.updateDeleteFlag(ninosToUndoDeletion, revertSoftDelete = true).value.onComplete {
+      case Success(Right(undeletedAccounts)) =>
+        val ninos = undeletedAccounts.map(_.nino).toList
```
lutnos, line 59: Ditto

lutnos (changes requested):
You just need to actually audit the document IDs, not just the NINO, please

mahesh-hmrc:
> You just need to actually audit the document IDs, not just the NINO, please

Yeah, sorted now. On back of our chat to revert on config, I accidentally reverted more than I should have. Included the code back in to audit doc id aswell. However, as part of that, felt it's clean to just keep the interface to NINODeletionConfig for ApplicationStart. Not ideal to change visibility of domain object outside of current `repo` scope

lutnos (approved):

lutnos (approved):

PR #269, opened 2023-12-15
DLS-9402 | Add list of ninos to be marked for soft deletion

mahesh-hmrc:


LAJW (changes requested):
Shouldn't we add the ninos to app-config repository instead?

mahesh-hmrc:
Config is now in appconfig-production as it should have been. Closing this as it's handled as part of https://github.com/hmrc/app-config-production/pull/20680

PR #273, opened 2024-02-27
DLS-10074: Play 30 migration.

BetaDraconis:


karolina-zaborska (changes requested):

project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.9.7
+sbt.version=1.9.9
```
karolina-zaborska, line 1: I'm just curious is it right to use version 1.9.9 of sbt. You are probably right but I'm commenting that just to double-check

image

karolina-zaborska (approved):

PR #279, opened 2024-05-17
DLS-9623: Removed PagerDuty alert

jackwillett1:


app/uk/gov/hmrc/helptosave/services/HelpToSaveService.scala
```diff
@@ -157,7 +157,6 @@ class HelpToSaveServiceImpl @Inject()(
                 logger.warn(
                   s"$tag Could not parse JSON response from paye-personal-details, received 200 (OK): $e ${timeString(time)}",
                   nino, params(response))
-                pagerDutyAlerting.alert("Could not parse JSON in the paye-personal-details response")
```
karolina-zaborska, line 160: just out of curiosity, why did we remove that?

jackwillett1: It was basically decided that this is working as expected so we just want to stop getting the PagerDuty alerts. You can find further details in Lukasz's answers in the Slack thread here https://hmrcdigital.slack.com/archives/C03D5C9F6RX/p1715864527408199

karolina-zaborska (approved):

PR #281, opened 2024-05-21
[DLS-6490] amend log

ruthdas03:


app/uk/gov/hmrc/helptosave/controllers/PayePersonalDetailsController.scala
```diff
@@ -45,7 +45,7 @@ class PayePersonalDetailsController @Inject()(
       .getPersonalDetails(nino)
       .fold(
         { error =>
-          logger.warn(s"Could not retrieve paye-personal-details from DES: $error", nino)
+          logger.warn(s"Could not retrieve paye-personal-details: $error", nino)
```
TomRafferty, line 48: Do we want to mention IF here instead of DES?
Or is that included in `error` ?

ruthdas03: So this is the error for calling the payeDetails method for DES or IF and we log it in their respective called methods, where we are logging if it's for DES or IF

TomRafferty: perfect, thanks for that

PR #287, opened 2024-10-11
[DLS-10533] migrating from http client to http client v2

ruthdas03:


test/uk/gov/hmrc/helptosave/modules/UCThresholdOrchestratorSpec.scala
```diff
@@ -62,7 +62,7 @@ class UCThresholdOrchestratorSpec extends ActorTestSupport("UCThresholdOrchestra
 
       connector
         .getThreshold()(*, *)
-        .returns(Future.successful(HttpResponse(500, "")))
+        .returns(Future.successful(Right(HttpResponse(500, ""))))
```
mahesh-hmrc, line 65: This doesn't seem rite. With the new error handling in HttpClientV2, we will never get a right based HttpResponse for 4xx and 5xx. I would check that the connector function has correct return type and that we handle this as a Left on the error string or whatever is relevant to check here.

mahesh-hmrc:
Also, now that they are resolved, can we remove the test failure descriptions from the PR comment. Thanks

ruthdas03:
> Also, now that they are resolved, can we remove the test failure descriptions from the PR comment. Thanks

Removed them

mahesh-hmrc (dismissed):

app/uk/gov/hmrc/helptosave/connectors/BarsConnector.scala
```diff
@@ -36,18 +37,19 @@ trait BarsConnector {
...
 
-  private val headers = Map("Content-Type" -> "application/json")
+  private val headers:(String, String) = "Content-Type" -> "application/json"
 
   override def validate(request: BankDetailsValidationRequest, trackingId: UUID)(
     implicit hc: HeaderCarrier,
-    ec: ExecutionContext): Future[HttpResponse] =
-    http.post(barsEndpoint, bodyJson(request), headers.+("X-Tracking-Id" -> trackingId.toString))
+    ec: ExecutionContext): Future[HttpResponse] = {
+    http.post(barsEndpoint).withBody(bodyJson(request)).transform(_.addHttpHeaders(headers, "X-Tracking-Id" -> trackingId.toString)).execute[HttpResponse]
```
mahesh-hmrc, line 51: Sorry, missed this in previous review. 
Would it also make sense to capture the return type as `Either[UpstreamErrorResponse, HttpResponse]` here? This might need tests being tidied up to reflect the change

While doing that, could you also check on other connectors making sure of the return type and error handling in calling code.

mahesh-hmrc (approved):
looks good, thanks

mahesh-hmrc (approved):

PR #298, opened 2025-07-28
Shivangi/DLS-11996 : Scala 3.71.1 upgrade

ShivangiShah42:


app/uk/gov/hmrc/helptosave/controllers/EnrolmentStoreController.scala
```diff
@@ -83,23 +83,23 @@ class EnrolmentStoreController @Inject() (
...
-      e => InternalServerError,
+      _ => InternalServerError,
       identity
     )
 
   private def processGetAccountNumberFromNSI(nino: NINO, uri: String)(implicit hc: HeaderCarrier): Future[Result] =
     getAccountNumberFromNSI(nino, uri).fold(
       { e =>
-        logger.info("Call to getAccountNumberFromNSI returned error response")
+        logger.info(s"Call to getAccountNumberFromNSI returned error response:$e")
```
gajalakshmiseetharaman: Concern about displaying exception in logs

mahesh-hmrc: Thanks @gajalakshmiseetharaman. Yeah, that's a valid concern and we shouldn't be logging the exception detail for third party integrations as we never know what detail they will be including in them. For ex, they could include NINOs in the error that we will end up accidentally logging them at INFO level.

Best is to leave as is or change log level to `debug` (which doesn't give us much of a benefit considering the application log level). So, I would just keep the way it was where we don't include exception details there.

gajalakshmiseetharaman (approved):

